PR #11, opened 2017-03-13
Tool refactored to externalize all configuration files.

### LINKED ISSUE #7: Move the configuration of the checks out of the Maven plugin
The configuration of the checks shouldn't be included in the Maven plugin. At the moment it is stored in the ''src/main/resources/" folder, but should be moved to the repository that is using the plugin.

This will allow having a different configurations for the same check, when using the tool in different builds.

svilenvul:
Several changes applied:
- the tool will keep a single set of configuration files, which will be
  used by default;
- added new parameters to the MOJOs that will allow adding external
  configuration files;
- FindBugs detectors are moved to an XML file;
- update the documentation;
- update the tests.

Fixes #7


kaikreuzer:
Thanks! I'll do a 0.0.2 release with this now and adapt the openhab2-addons build accordingly.

Issue #23, opened 2017-03-20
Add some documentation about covered guidelines


kaikreuzer:
@svilenvul I am personally not having the overview about which [bullets from our guidelines](https://www.eclipse.org/smarthome/documentation/development/guidelines.html) are already covered by the tool and which aren't yet.

Would you be able to provide a short info about that in the tool's documentation / README?

This would also be very helpful to track, which new checks might still need to be implemented.

petarnv:
@kaikreuzer, I have opened a PR for adding the coverage of the `ESH Coding Guidelines` in the documentation of the tool(https://github.com/openhab/static-code-analysis/pull/24). Maybe it is good to be noticed that there are checks not only from these guidelines. All the included checks at the moment from `checkstyle`, `findbugs `and `pmd `can be seen in the rulesets (https://github.com/openhab/static-code-analysis/tree/master/src/main/resources/rulesets).

kaikreuzer:
Thanks @petarnv!
Would it make sense to add also information on whether it is a "strict" check, i.e. it makes the build fail? This is something that I would expect for most (if not all) of the checks against the guideline.

> Maybe it is good to be noticed that there are checks not only from these guidelines.

Correct, but only a few ones from findbugs and PMD will be considered as errors and break the build, right? Actually, if those are a pre-requisite for having a PR merged, we should add those checks to the guidelines section, wdyt? (Could be a single new entry, referencing the pmd/findbug problem descriptions).

petarnv:
> Would it make sense to add also information on whether it is a "strict" check

Yes, I will update my PR with information about whether a check is "strict".

> only a few ones from findbugs and PMD will be considered as errors and break the build, right? 

I don't think that any check from findbugs and PMD should break the build.

> we should add those checks to the guidelines section, wdyt?

I think that for now the guidelines are fine. If we find another check, that should break the build, maybe then we can discuss changing the guidelines section.

kaikreuzer:
> I don't think that any check from findbugs and PMD should break the build.

Well, this is how it is today, see the "Usage" documentation:

```
for PMD - the build will fail when a rule with Priority "1" is found;
for Checkstyle - a rule with severity="Error";
for Findbugs - any Matcher with Rank between 1 and 4.
```

And there are [PMD rules with priority 1](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/pmd/rules.xml#L79) in our checks.

svilenvul:
[...]

Issue #12, opened 2017-03-13
Add check for tabs in java files


kaikreuzer:
Our code formatter is using spaces and no tabs in java files.
I'd therefore claim that any tab within a java file is not correct and should be marked as an error.

Could such a check be added?

svilenvul:
Yes, this is the [FileTabCharacterCheck](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L17) and is already included

From the javadoc:
>Checks to see if a file contains a tab character.

The tool found a tab in the [AutelisBindingConstants](https://github.com/openhab/openhab2-addons/blob/50f7f662b4e5cf9f99b0c9a3e64f7097357d17e4/addons/binding/org.openhab.binding.autelis/src/main/java/org/openhab/binding/autelis/AutelisBindingConstants.java) and other classes in the autelis binding.

kaikreuzer:
Ah, cool, thanks.

kaikreuzer:
> ...and should be marked as an error.

This seems to be currently "priority 3", which I assume will not result in a failing build, right?
Is it possible to change the priority of single checks, i.e. increase this one to be an error?

svilenvul:
Yes it can be changed and only priority 1 will fail the build.

Actually we haven't set any priority 1 issues now (only bugs found by FindBugs will fail the build now).

kaikreuzer:
> Actually we haven't set any priority 1 issues now

So we should start with it.
Imho, everything that makes me ask for changes on a PR is also a candidate to fail the build (at least where no false positives can be expected).

And missing code formatter is definitely such a case.

> OK, I will go for it and prepare a PR for openhab2-addons too.

imho this can be part of the default rules, because the very same applies to ESH.

svilenvul:
> OK, I will go for it and prepare a PR for openhab2-addons too

What I meant to say was that all tabs should be removed from the .java files in openhab2-addons in order to keep the build green.

kaikreuzer:
Ah, right, even better :-)

PR #24, opened 2017-03-20
Added documentation about the coverage of ESH Coding Guidelines

petarnv:
I saw this opened issue - https://github.com/openhab/static-code-analysis/issues/23. I have added an `ESH Guidelines Covered` section to the README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,12 +227,15 @@ For more information: https://pmd.github.io/pmd-5.4.1/customizing/rule-guideline
 # ESH Guidelines Covered
 
 ## A. Code Style
-1. The Java naming conventions should be used. - https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L80
-2. Every Java file must have a license header. You can run mvn license:format on the root of the repo to automatically add missing headers. - https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L34
-3. Every class, interface and enumeration should have JavaDoc describing its purpose and usage. - https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L66
+1. [The Java naming conventions should be used.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L80) - `severity=info`
+2. [Every Java file must have a license header. You can run mvn license:format on the root of the repo to automatically add missing headers.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L34) - `severity=info`
+3. [Every class, interface and enumeration should have JavaDoc describing its purpose and usage.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L66) - `severity=warning`
 4. Every class, interface and enumeration must have an @author tag in its JavaDoc for every author that wrote a substantial part of the file. - `Work in Progress`
 5. Every constant, field and method with default, protected or public visibility should have JavaDoc (optional, but encouraged for private visibility as well). - `Not covered yet`
-6. Code must be formatted using the provided code formatter and clean up settings. They are set up automatically by the official IDE setup. - `Not covered yet`
+6. The code must be formatted:
+    - [java files must use spaces for formatting, rather than tabs](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L17)
+    - xml files must use tabs for formatting, rather than spaces - `Not covered yet`
+    - json files must use tabs for formatting, rather than spaces - `Not covered yet`
 7. Generics must be used where applicable. - `Not covered yet`
 8. Code should not show any warnings. Warnings that cannot be circumvented should be suppressed by using the @SuppressWarnings annotation. - `Not covered yet`
 9. For dependency injection, OSGi Declarative Services should be used. - `Work in Progress`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -224,7 +224,49 @@ For more information: https://pmd.github.io/pmd-5.4.1/customizing/rule-guideline
 
 - Flooded console output when running Checkstyle in debug mode in Maven  (- X ) - https://github.com/checkstyle/checkstyle/issues/3184;
 
+# ESH Guidelines Covered
+
+## A. Code Style
+1. The Java naming conventions should be used. - https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L80
```
kaikreuzer: Maybe better make this to a link for better readability: 
```
1. [The Java naming conventions should be used.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L80)
```

README.md
```diff
@@ -224,7 +224,49 @@ For more information: https://pmd.github.io/pmd-5.4.1/customizing/rule-guideline
...
+2. Every Java file must have a license header. You can run mvn license:format on the root of the repo to automatically add missing headers. - https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L34
+3. Every class, interface and enumeration should have JavaDoc describing its purpose and usage. - https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L66
+4. Every class, interface and enumeration must have an @author tag in its JavaDoc for every author that wrote a substantial part of the file. - `Work in Progress`
+5. Every constant, field and method with default, protected or public visibility should have JavaDoc (optional, but encouraged for private visibility as well). - `Not covered yet`
+6. Code must be formatted using the provided code formatter and clean up settings. They are set up automatically by the official IDE setup. - `Not covered yet`
```
kaikreuzer: This should probably be split up - we have dozens of code formatter settings, and we won't (and don't want) to maintain them separately here as well in all details.

With #12, we will have the most common issue covered already - we should add a similar check for xml and json files (where we actually expect tabs, no spaces). Having these checks in place, I would consider A.6 to be checked well enough.

README.md
```diff
@@ -224,7 +224,49 @@ For more information: https://pmd.github.io/pmd-5.4.1/customizing/rule-guideline
...
+
+## B. OSGi Bundles
+1. Every bundle must contain a Maven pom.xml with a version and artifact name that is in sync with the manifest entry. The pom.xml must reference the correct parent pom (which is usually in the parent folder). - `Work in Progress`
+2. Every bundle must contain an about.html file, providing license information. - `Work in Progress`
+3. Every bundle must contain a build.properties file, which lists all resources that should end up in the binary under bin.includes. - `Work in Progress`
+4. The manifest must not contain any “Require-Bundle” entries. Instead, “Import-Package” must be used. - `Opened PR` (https://github.com/openhab/static-code-analysis/pull/19)
+5. The manifest must not export any internal package. - https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L40
+6. The manifest must not have any version constraint on package imports, unless this is thoughtfully added. Note that Eclipse automatically adds these constraints based on the version in the target platform, which might be too high in many cases. - `Work in Progress`
```
maggu2810: > The manifest must not have any version constraint on package imports, unless this is thoughtfully added.

How do you plan to check if a version constraint is added thoughtfully? :wink:

README.md
```diff
@@ -224,7 +224,49 @@ For more information: https://pmd.github.io/pmd-5.4.1/customizing/rule-guideline
...
+7. Generics must be used where applicable. - `Not covered yet`
+8. Code should not show any warnings. Warnings that cannot be circumvented should be suppressed by using the @SuppressWarnings annotation. - `Not covered yet`
+9. For dependency injection, OSGi Declarative Services should be used. - `Work in Progress`
+10. Packages that contain classes that are not meant to be used by other bundles should have “internal” in their package name. - `Work in Progress`
+
+## B. OSGi Bundles
+1. Every bundle must contain a Maven pom.xml with a version and artifact name that is in sync with the manifest entry. The pom.xml must reference the correct parent pom (which is usually in the parent folder). - `Work in Progress`
+2. Every bundle must contain an about.html file, providing license information. - `Work in Progress`
+3. Every bundle must contain a build.properties file, which lists all resources that should end up in the binary under bin.includes. - `Work in Progress`
+4. The manifest must not contain any “Require-Bundle” entries. Instead, “Import-Package” must be used. - `Opened PR` (https://github.com/openhab/static-code-analysis/pull/19)
```
maggu2810: :+1: 
But I assume this could be a `hard work` for the model bundles.

Should we exclude test bundles?

README.md
petarnv: Well, we can add configuration properties for packages, that are usually imported with a version (like `org.junit`, `org.apache.commons`, `org.slf4j` and probably others) and these packages will not be considered a violation, even if they are imported with a version.

Also, if there is a bundle, that really needs to import a package with a version, it can be added to the suppressions.xml file (https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/suppressions.xml) and it will not be checked.

README.md
petarnv: Yes, we can exclude the model bundles. But I'm not sure what do you mean when you say that checking the model bundles will be hard work?

README.md
petarnv: Done.

README.md
petarnv: It is now split on three parts - formatting for java, xml and json.

petarnv:
I updated the PR. Now for every implemented check, its severity/priority can be seen. 

I have also added a section for all checks that we have from Checkstyle, PMD and Findbugs with links to their configuration files and their documentation.

README.md
kaikreuzer: @maggu2810 For the model bundles, we should be able to deactivate this check in the pom.xml (@petarnv, I understood that this is possible for individual checks, right?)

PR #25, opened 2017-03-21
Implemented RequiredFilesCheck and tests for it.

petarnv:
The check verifies if the required files for the bundle are present. These required files are listed as a configuration properties. At this initial version as a required files are considered about.html, build.properties, pom.xml, MANIFEST.MF and README.md. 
Also, I have implemented tests for the check.

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/RequiredFilesCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/RequiredFilesCheck.java
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.openhab.tools.analysis.checkstyle.api.AbstractStaticCheck;
+
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+import com.puppycrawl.tools.checkstyle.api.MessageDispatcher;
+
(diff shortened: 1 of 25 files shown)


petarnv:
@kaikreuzer, there are a lot of bindings that broke the build, because they don't have `about.html` files. Would you like me to make a PR for adding `about.html`'s in these bindings

kaikreuzer:
I'd assume that these are the ones that don't have any 3rd party libs included (and on those I didn't enforce that in the past).
But yes, it would be nice if you could add it - as it should be the same file for all of them, I hope it won't mean much work for you.

petarnv:
@kaikreuzer, I am preparing a PR for adding the missing about.html files, but I have a question. Do I need to add about.html files in the test bundles?

Also, a lot of the test bundles don't have a README.md, which I guess is normal, but it breaks the build. And that brings another question. Should we suppress the test bundles and not check whether they contain the required files?

petarnv:
> @kaikreuzer, I am preparing a PR for adding the missing about.html files, but I have a question. Do I need to add about.html files in the test bundles?

I made a PR for adding about.html files where missing - https://github.com/openhab/openhab2-addons/pull/2065. For now I have added them in the test bundles, too. But if they don't have to be added there, I can remove them from the PR.

kaikreuzer:
> Should we suppress the test bundles and not check whether they contain the required files?

Yes, I think we should exclude them from the checks as quite some different requirements might apply on them.

petarnv:
I updated this PR and now the test bundles are excluded. 

However, there are still two bundles, that fail, because they don't have README.md files:
```
[INFO] Transport Feed bundle .............................. FAILURE [  0.531 s]
[INFO] PHP Support for the CometVisu Backend .............. FAILURE [  7.222 s]
```
@kaikreuzer, what do you think - should we open an issue for these bundles not having README.md files, or should we exclude them, too?

kaikreuzer:
Afaik, we only defined that README.mds are mandatory for bindings, but not for any OSGi bundle - so imho it is correct that they are missing here. Could the check be weakened accordingly?

petarnv:
I will update this PR when https://github.com/openhab/static-code-analysis/issues/31 is resolved.

PR #30, opened 2017-03-24
Add a check for bundle vendors 

martinvw:
Add a check for proper bundle vendor, I am unsure whether the openHAB repo should contain bundles with eclipse smarthome but it does at this moment (org.openhab.binding.urtsi). 

If this test is NOT useful please let me know as well, thanks!

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle;
+
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+import org.openhab.tools.analysis.checkstyle.api.AbstractStaticCheck;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Checks if a manifest file contains the expected bundle vendor
+ *
+ * @author Martin van Wingerden
+ */
+public class BundleVendorCheck extends AbstractStaticCheck {
+    private final static String MANIFEST_EXTENSION = "MF";
+    private static final String REQUIRED_PREFIX = "Bundle-Vendor: ";
+    private List<String> allowedValues;
+
+    public BundleVendorCheck() {
+        setFileExtensions(MANIFEST_EXTENSION);
+    }
+
+    public void setAllowedValues(String[] allowedValues) {
+        this.allowedValues = Arrays.asList(allowedValues);
+    }
+
+    @Override
+    protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
+        if (isEmpty(file)) {
+            // not our task to report
+            return;
+        }
+
+        List<String> bundleVendors = lines.stream()
+                .filter(line -> line.toLowerCase().startsWith("bundle-vendor"))
+                .collect(Collectors.toList());
+
+        boolean tooMany = false;
+        if (bundleVendors.size() == 0) {
+            log(0, "\"Bundle-Vendor\" is missing");
+            return;
+        } else if (bundleVendors.size() > 1) {
+            tooMany = true;
+        }
+
+        int lineNumber = 0;
+        for (String bundleVendor : bundleVendors) {
+            lineNumber = findLineNumber(lines, bundleVendor, lineNumber);
+
+            if (tooMany) {
+                log(lineNumber, "Only one \"Bundle-Vendor\" was expected.");
+            }
+            if (!bundleVendor.startsWith(REQUIRED_PREFIX)) {
+                log(lineNumber, "Expect eg. \"Bundle-Vendor: openHAB\" got \"" + bundleVendor + "\"");
+            } else {
+                String onlyValue = bundleVendor.replace(REQUIRED_PREFIX, "");
+                if (!allowedValues.contains(onlyValue)) {
+                    log(lineNumber, "Unexpected \"" + onlyValue + "\", only allowed options: " + allowedValues);
+                }
+            }
+        }
+    }
+}
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
@@ -24,18 +24,18 @@ import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
  */
 public class ExportInternalPackageCheck extends AbstractStaticCheck {
 
-    public final static String MANIFEST_EXTENSTION = "MF";
+    public final static String MANIFEST_EXTENSION = "MF";
     public final static String MESSAGE_INTERNAL_PACKAGE_EXPORTED = "Remove internal package export";
     public final static String MESSAGE_FILE_EMPTY = "File is empty !";
 
     public ExportInternalPackageCheck() {
-        setFileExtensions(MANIFEST_EXTENSTION);
+        setFileExtensions(MANIFEST_EXTENSION);
     }
 
     @Override
     protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
         if (isEmpty(file)) {
-            log(0, MESSAGE_FILE_EMPTY, new Integer(0));
+            log(0, MESSAGE_FILE_EMPTY, 0);
             return;
         }
         BundleInfo manifest = parseManifestFromFile(file);
@@ -43,10 +43,10 @@ public class ExportInternalPackageCheck extends AbstractStaticCheck {
 
         int lineNumber = 0;
         for (Object export : exports) {
-            String pacakgeName = export.toString();
-            if (pacakgeName.contains(".internal")) {
-                lineNumber = findLineNumber(lines, pacakgeName, lineNumber);
-                log(lineNumber, MESSAGE_INTERNAL_PACKAGE_EXPORTED + " " + pacakgeName, new Integer(0));
+            String packageName = export.toString();
+            if (packageName.contains(".internal")) {
+                lineNumber = findLineNumber(lines, packageName, lineNumber);
+                log(lineNumber, MESSAGE_INTERNAL_PACKAGE_EXPORTED + " " + packageName, 0);
             }
         }
 
--- a/src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
+++ b/src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
@@ -1,0 +1,101 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle.test;
+
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.TreeWalker;
+import com.puppycrawl.tools.checkstyle.api.Configuration;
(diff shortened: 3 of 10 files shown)


martinvw:
I did not yet link it to the configuration but lets first discuss whether it is indeed useful :-)

src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
```diff
@@ -24,29 +24,29 @@
  */
 public class ExportInternalPackageCheck extends AbstractStaticCheck {
 
-    public final static String MANIFEST_EXTENSTION = "MF";
+    public final static String MANIFEST_EXTENSION = "MF";
```
dstivanov: I guess that you have looked into this class to write your new check - and thus you have applied some formatting on it. Can you please  revert it back to make the PR clear.

src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
```diff
@@ -24,29 +24,29 @@
...
     public ExportInternalPackageCheck() {
-        setFileExtensions(MANIFEST_EXTENSTION);
+        setFileExtensions(MANIFEST_EXTENSION);
     }
 
     @Override
     protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
         if (isEmpty(file)) {
-            log(0, MESSAGE_FILE_EMPTY, new Integer(0));
+            log(0, MESSAGE_FILE_EMPTY, 0);
```
dstivanov: I wonder why this was Integer.. Probably @svilenvul can explain it as an author of the check.

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
```diff
@@ -0,0 +1,68 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
```
petarnv: I think that the license headers in openHab are now updated to 2017: https://github.com/openhab/openhab2-addons/pull/1865. Can you please change it, so that we follow the same pattern?

src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Tests for {@link BundleVendorCheck}
+ *
+ * @author Martin van Wingerden
+ */
+public class BundleVendorCheckTest extends AbstractStaticCheckTest {
+    private static final String TEST_DIRECTORY = "correctOpenHABSpellingInManifestTest/";
+    private DefaultConfiguration checkConfig;
+
+    @Before
```
petarnv: Don't you use the same configuration for every test? If that is the case can you please make the annotation of the method `@BeforeClass`, so that we create this configuration just one time?

src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
svilenvul: This is one of the first checks that we wrote and we wrote it against the Google checkstyle API (which is slightly different and I think it accepted only objects).

src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
```diff
@@ -0,0 +1,94 @@
+package org.openhab.tools.analysis.checkstyle.test;
```
svilenvul: Please add a header

src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
```diff
@@ -0,0 +1,94 @@
...
+        );
+    }
+
+    private void verify(String fileName, String[] expectedMessages) throws Exception {
+        verify(checkConfig,
+                getPath(TEST_DIRECTORY + fileName),
+                expectedMessages
+        );
+    }
+}
```
svilenvul: Please add a new line

svilenvul:
@kaikreuzer we have a PR for check that validates the Bundle-Vendor manifest header.

Maybe I miss something, but I can't find any information in the guidelines. Should/Must the Bundle-Vendor  be "openHAB", or any third party organizations could add their name? Could we include this check?

src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
martinvw: I corrected spelling errors, do you want a separate PR for it?

src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
martinvw: Since Java 5 autoboxing exists, so it is unlikely that it was required...

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
martinvw: Yes of course, I copied my class, we need a PR for all other classes as well :-)

src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
martinvw: The test which I saw in the existing code constructed it in each test so I used the same pattern but using an @Before. IMHO it's not that important if it's @Before or @BeforeClass so I can change it.

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
```diff
@@ -0,0 +1,68 @@
...
+        } else if (bundleVendors.size() > 1) {
+            tooMany = true;
+        }
+
+        int lineNumber = 0;
+        for (String bundleVendor : bundleVendors) {
+            lineNumber = findLineNumber(lines, bundleVendor, lineNumber);
+
+            if (tooMany) {
+                log(lineNumber, "Only one \"Bundle-Vendor\" was expected.");
```
martinvw: It requires way more String objects (because you have to compose them) and imho it will not become more readable...

Example `"Only one \"Bundle-Vendor\" was expected."`

would become:  
``` 
"Only one \"" +BUNDLE_VENDOR + "\" was expected.";
```
 Would lead to 3 additional String objects (although the compiler might resolve this for) while gaining nothing in readability.

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Checks if a manifest file contains the expected bundle vendor
+ *
+ * @author Martin van Wingerden
+ */
+public class BundleVendorCheck extends AbstractStaticCheck {
+    private final static String MANIFEST_EXTENSION = "MF";
+    private static final String REQUIRED_PREFIX = "Bundle-Vendor: ";
+    private static List<String> ALLOWED_VALUES = Arrays.asList("Eclipse.org/SmartHome", "openHAB");
```
martinvw: Great, do you already have a class with an example for that?

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
```diff
@@ -0,0 +1,68 @@
...
+
+        int lineNumber = 0;
+        for (String bundleVendor : bundleVendors) {
+            lineNumber = findLineNumber(lines, bundleVendor, lineNumber);
+
+            if (tooMany) {
+                log(lineNumber, "Only one \"Bundle-Vendor\" was expected.");
+            }
+            if (!bundleVendor.startsWith(REQUIRED_PREFIX)) {
+                log(lineNumber, "Expect eg. \"Bundle-Vendor: openHAB\" got \"" + bundleVendor + "\"");
```
martinvw: This specific error is about incorrect capitalization of 'Bundle-Vendor', missing colons, spaces things like that. So it looks like Bundle-Vendor but is not starting with "Bundle-Vendor: "

martinvw: Reusing the strings to the test does imho make the tests less readable and if you somehow accidently empty or otherwise mutilate your string your test remain happy. 

If we really want that can we then please put the tests into the same package so those strings do not have to be public.

src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
martinvw: Will do, testing this is the goal of this project :-S

src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
martinvw: Will do, testing this is the goal of this project :-S

martinvw:
Thanks for the feedback I have some replies on things please let me know what is important for you to accept it. I will definitely fix the style errors.

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
petarnv: Yes, some of the other classes need to be updated, too. :) 
Thanks for noticing.

src/test/java/org/openhab/tools/analysis/checkstyle/test/BundleVendorCheckTest.java
petarnv: Whatever you decide.

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
dstivanov: Yes! not yet merged, but present in a PR. See here: https://github.com/openhab/static-code-analysis/pull/25/files#diff-82672bd78945c746d7c6386e8df15080R45

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
dstivanov: Your argument is good, so I agree. leave it like that :)

src/main/java/org/openhab/tools/analysis/checkstyle/BundleVendorCheck.java
dstivanov: I got your point, in the way it is now we will have passing tests even if we have incorrect messages. So yes, it is better to leave the tests with hardcoded messages,

src/main/java/org/openhab/tools/analysis/checkstyle/ExportInternalPackageCheck.java
dstivanov: No, I don't want separate PR, I just missed the spelling error. Thanks for fixing them!

kaikreuzer:
> Should/Must the Bundle-Vendor be "openHAB", or any third party organizations could add their name? Could we include this check?

Yes, this check makes a lot of sense. Our archetype creates it as "openHAB" and it must not be changed for any official openHAB bundles.

martinvw:
The is also one binding in the openhab project which has eclipse smarthome as bundle vendor, should this be fixed? See text of the pr.

Should i also include (some) other default from the archetype, if yes, which?

I'll work on the comments later so that is ready to be merged.

kaikreuzer:
> The is also one binding in the openhab project which has eclipse smarthome as bundle vendor, should this be fixed?

Yes, please.

> Should i also include (some) other default from the archetype, if yes, which?

I don't see anything specific to add.

martinvw:
All done:

- added the missing file header
- updated the year in the copyright header 
- added a new line at the end of the file
- made the allowed values configurable 
- use a BeforeClass instead of normal Before.
- created a PR for the OH2 binding with the wrong vendor

IMHO all should be fine now.

martinvw:
@kaikreuzer should the check for Java 8 also be included in this check it seems like a logical place for it? 

The class could then use some renaming but that shouldn't be a problem.

dstivanov:
@martinvw We are having several more checks for the Manifest as well, so I guess that the Java version can be added into separate check. By having the checks more distributed we will have the freedom to control whether they are executed or not, without having to alter the code, but only their configurations.

svilenvul:
Thanks !

PR #32, opened 2017-03-24
Changed the year in the license headers from 2016 to 2017

petarnv:
As @martinvw noticed in https://github.com/openhab/static-code-analysis/pull/30 our license headers are not updated with the correct year as it is already done in openhab2-addons (https://github.com/openhab/openhab2-addons/pull/1865) and I think it is good to follow the same pattern.


svilenvul:
Thanks !

PR #46, opened 2017-04-02
Add a check for modifier order

martinvw:
Since at has just been fixed in openHAB it would be nice to add the check too

See PR:
https://github.com/openhab/openhab2-addons/pull/2098
https://github.com/openhab/openhab2-addons/pull/2113

http://checkstyle.sourceforge.net/config_modifier.html#ModifierOrder


svilenvul:
Thanks !

PR #52, opened 2017-04-05
A check for author tag is implemented. Some tests for it are added.

m-memova:
Signed-off-by: Mihaela Memova <mihaela.memova@musala.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ For more information: https://pmd.github.io/pmd-5.4.1/customizing/rule-guideline
 1. [The Java naming conventions should be used.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L80) - `severity=info`
 2. [Every Java file must have a license header. You can run mvn license:format on the root of the repo to automatically add missing headers.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L34) - `severity=info`
 3. [Every class, interface and enumeration should have JavaDoc describing its purpose and usage.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L66) - `severity=warning`
-4. Every class, interface and enumeration must have an @author tag in its JavaDoc for every author that wrote a substantial part of the file. - `Work in Progress`
+4. [Every class, interface and enumeration must have an @author tag in its JavaDoc for every author that wrote a substantial part of the file.](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml#L97) - `severity=error`
 5. Every constant, field and method with default, protected or public visibility should have JavaDoc (optional, but encouraged for private visibility as well). - `Not covered yet`
 6. The code must be formatted:
     - java files must use spaces for formatting, rather than tabs - `severity=error`;
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/AuthorTagCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/AuthorTagCheck.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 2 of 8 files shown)


svilenvul:
Great, thanks, please update the README.MD file with the priority of the check.

m-memova:
@kaikreuzer , are you OK with this check failing the build or should I change its priority to "warning"?

kaikreuzer:
Failing the build is fine, thanks!

m-memova:
Done! :)

m-memova:
So, I tried this check in the addons2 repository. And it turned out that more than 100 classes have no author tags. That means that when this PR is merged, the build will fail for every new PR. So I searched through all the bindings and I made a list of potentials authors. I think it would be nice to make an issue [here](https://github.com/openhab/openhab2-addons/issues). If the authors add their tags before merging this PR, the build will be successful. @kaikreuzer, @svilenvul , what do you think?

kaikreuzer:
I fear that it will take forever to ask the individual authors to add that.
Question: Is it easily possible to deactivate certain rules on certain bundles? Then we could add an "ignore" to the pom.xmls of the concerned bundles and still would have the check in place for any new contribution.

m-memova:
I can add all the files-with-no-author tag in the  suppressions.xml file. But I am not sure that would be a better decision. Besides, there are several things to consider : 
- 71 of those files are from org.openhab.ui.cometvisu (judging by the rest of the files there, we can be pretty sure that the author is Tobias Bräutigam)
- many of the files with no author tag are from internal packages (according to the ESH guidelines every file must have an author tag but still...)

kaikreuzer:
> 71 of those files are from org.openhab.ui.cometvisu (judging by the rest of the files there, we can be pretty sure that the author is Tobias Bräutigam)

I just quickly scanned through this bundle and only saw a single class where the tag was missing. Do you have details?

> many of the files with no author tag are from internal packages (according to the ESH guidelines every file must have an author tag but still...)

There might be SOME exceptions, like e.g. generated code (e.g. JAXB beans) or third-party code that has been included in the repo - both are valid exceptions to the rule, so for those cases it would be nice if there is a chance to configure the bundle in a way that this check is suppressed.

martinvw:
@kaikreuzer we can of course track back the authors by doing some git blames, or is that not sufficient.

kaikreuzer:
@martinvw That won't help for the two exceptional cases that I mentioned (generated & 3rd party code).

m-memova:
@kaikreuzer , you are right, 70 of the files with no author tag in org.openhab.ui.cometvisu are generated by the JavaTM Architecture. I will think about how we can handle those exceptions of the rule.

martinvw:
In most projects I worked on professionally we put the generated code in the target directory and let maven generate it on each build, we could try whether that is a possible option here too. It would solve this problem and make sure the generated code is not accidentally altered 

@kaikreuzer what would you think about this?

svilenvul:
I think for now we can just add them to the suppression filter. (since most of the files are in a single bundle).
@m-memova, what do you think about this ?

kaikreuzer:
> we put the generated code in the target directory and let maven generate it on each build

I only do this for code that we do not compile against ourselves - the rest is checked in, because otherwise you will have hundreds of compilation errors after doing the IDE setup.
This is the beauty of the OH2 IDE setup over the ESH IDE: It does not require any additional code generation steps and thus is much simpler to get started with.

martinvw:
> This is the beauty of the OH2 IDE setup over the ESH IDE: It does not require any additional code generation steps and thus is much simpler to get started with.

Maybe that is more something to debate over beer, but I would not consider it beauty that you lose track of the difference between generated code and potentially customized (generated) code :-)

kaikreuzer:
Yes, I go for the beer 🍻

svilenvul:
I am sure that all of us would like to see this check merged :-), so I will start resolving some of the errors. I still have some questions :
- there are some bundles that contain one ore two classes that miss the author tag. Is it OK to add the authors using the information from git blame?
- there are some auto generated classes. Is it possible to move them to ``src/gen/java`` source folder and ignore them directly in the check, or they will be added in the exclude filter ?
- there are some multi line comments that should be changed to javadoc;

martinvw:
> there are some auto generated classes. Is it possible to move them to src/gen/java source folder and ignore them directly in the check, or they will be added in the exclude filter ?

I think that that is great solution

svilenvul:
The cleanup is done :).

Are we ready to merge this check as well ?

Issue #45, opened 2017-04-01
Maven compatibility issues 


syphr42:
The ESH code formatter sets the max line width to 120 while the LineLengthCheck configured in this tool for CheckStyle flags lines over 80 characters as priority 1.

* What is the desired max line length for ESH & openHAB?
* The formatter and the CheckStyle rule should be aligned.

kaikreuzer:
Thanks for noticing! The desired length is indeed 120.

svilenvul:
I am little bit surprised, as the LineLengthCheck is not included in the [checkstyle ruleset](https://github.com/openhab/static-code-analysis/blob/master/src/main/resources/rulesets/checkstyle/rules.xml). Maybe some FindBugs or PMD check is logging this, but I haven't seen a failing build.

Can you provide an example in which bundle and file have you received this warning ?

syphr42:
It looks like there is a different problem. I was trying to put together a PR to fix the line length and I realized what @svilenvul mentioned - line length is not one of the checks in the rules.xml.

So digging into this some more, I found that the rules.xml is not being used at all. Instead, Checkstyle is executing sun_checks.xml (appears to be the default). It looks like the settings from the plugin are not executing, but the the goals are - it runs Checkstyle, FindBugs, and PMD. 

Because the properties are not being used, it is dropping the checkstyle-result.xml in target/ rather than target/code-analysis/ so the merge fails and report.html is zero bytes. If I manually copy the checkstyle-result.xml to the code-analysis folder and run again without 'clean' it will generate the report, but it is of course using the default rules.

Originally I was executing as
```
mvn clean install -P check
```

But in response to @kaikreuzer responding to my forum post [here](https://community.openhab.org/t/static-code-analysis-tool/24489/45) I just did a normal

```
mvn clean install
```

and I got the same result.

I then checked out master from the openhab2-addons repo and ran 'mvn clean install' on the Chromecast binding with the same result:

```
$ mvn clean install
[INFO] Scanning for projects...
[INFO] Computing target platform for MavenProject: org.openhab.binding:org.openhab.binding.chromecast:2.1.0-SNAPSHOT @ /home/moyerg/workspaces/eclipse-openhab2/openhab2-addons/addons/binding/org.openhab.binding.chromecast/pom.xml
[INFO] Fetching p2.index from http://download.eclipse.org/smarthome/updates-stable/
[INFO] Adding repository http://download.eclipse.org/smarthome/updates-stable
[...]

PR #54, opened 2017-04-06
Corrected some minor issues

martinvw:
While working on fixing #45 I noted some small imperfections:
 - The pom-version is a final one and not a snapshot, openHAB also uses the rule that normally its a snapshot
 - The version for the maven compiler plugin was not specified
 - The maven plug-in annotations should have the scope provided
 - Fixed some typo's
 - Changed the explicit array to list copy to a more modern way
 - Since properties are not referring to the user properties anymore I renamed the constants

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>org.openhab.tools</groupId>
   <artifactId>static-code-analysis</artifactId>
-  <version>0.0.3</version>
+  <version>0.0.4-SNAPSHOT</version>
   <packaging>maven-plugin</packaging>
 
   <name>Static Code Analysis Tool</name>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -6,7 +6,7 @@
 
   <groupId>org.openhab.tools</groupId>
   <artifactId>static-code-analysis</artifactId>
-  <version>0.0.3</version>
+  <version>0.0.4-SNAPSHOT</version>
```
svilenvul: You have your argument for this change: 
>The pom-version is a final one and not a snapshot, openHAB also uses the rule that normally its a snapshot

I think this might be a little bit inconvenient in our case for the people that develop new checks (as the OH2 addons build is using final version in the build). How are they supposed to execute them on the OH2 repo before merge ?

kaikreuzer: You would have to change the sat.version in the OH2 pom.xml to the snapshot version as well then. Or you locally change this line here back for your testing.
In general, I agree with @martinvw that it is cleaner to have a snapshot on the latest master (I just didn't care about it as we do not have any CI plan set up for this).

martinvw: The 'problem' with having a normal version here is that your local maven folder contains a different 0.0.3 version than mine, which is normal for snapshots but unexpected for final versions. I can ofcourse revert this line if both of you do not completely agree it's not crucial for me, it just felt cleaner.

svilenvul: > You would have to change the sat.version in the OH2 pom.xml to the snapshot version as well then.

All fine than.

PR #57, opened 2017-04-06
A check for importing all exported packages + tests for it.

m-memova:
Signed-off-by: Mihaela Memova <mihaela.memova@musala.com>

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
@@ -1,0 +1,80 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle;
+
+import java.io.File;
+import java.io.IOException;
+import java.text.ParseException;
+import java.util.List;
+import java.util.Set;
+
+import org.apache.ivy.osgi.core.BundleRequirement;
+import org.apache.ivy.osgi.core.ExportPackage;
+import org.apache.ivy.osgi.core.ManifestParser;
+import org.openhab.tools.analysis.checkstyle.api.AbstractStaticCheck;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+
+/**
+ * Checks if all of the exported packages are imported by the bundle itself
+ *
+ * @author Mihaela Memova
+ */
+public class ImportExportedPackagesCheck extends AbstractStaticCheck {
+
+    public static final String MANIFEST_EXTENSTION = "MF";
+    public static final String EXPORT_PACKAGES_HEADER = "Export-Package:";
+
+    Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    public ImportExportedPackagesCheck() {
+        setFileExtensions(MANIFEST_EXTENSTION);
+    }
+
+    @Override
+    protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
+        int lineNumberBeforeExportsStart = findLineNumber(lines, EXPORT_PACKAGES_HEADER, 0) - 1;
+
+        try {
+            Set<ExportPackage> exports = ManifestParser.parseManifest(file).getExports();
+            Set<BundleRequirement> imports = ManifestParser.parseManifest(file).getImports();
+
+            for (ExportPackage export : exports) {
+                if (!isPackageImported(imports, export)) {
+                    int lineToLog = findLineNumber(lines, export.toString(), lineNumberBeforeExportsStart);
+                    log(lineToLog, String.format("The exported package: %s is not imported", export));
+                }
+            }
+        } catch (IOException e) {
+            logger.error("An error occured while processing the file {}", file.getPath(), e);
+        } catch (ParseException e) {
+            logger.error("An error occured while trying to parse the MANIFEST {}", file.getPath(), e);
+        }
+    }
+
+    /**
+     * Checks if a particular package is imported. The default implementation of
+     * {@link Set#contains(Object)} cannot be used since an {@link ExportPackage} object
+     * is searched in a set of {link {@link BundleRequirement} objects.
+     *
+     * @param imports - set of all imported packages
+     * @param searchedPackage - the package that has to be checked
+     * @return true if the imports contains the package, false otherwise
+     */
+    private boolean isPackageImported(Set<BundleRequirement> imports, ExportPackage searchedPackage) {
+        for (Object o : imports) {
+            if ((o.toString()).equals(searchedPackage.toString())) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
--- a/src/test/java/org/openhab/tools/analysis/checkstyle/test/ImportExportedPackagesCheckTest.java
+++ b/src/test/java/org/openhab/tools/analysis/checkstyle/test/ImportExportedPackagesCheckTest.java
@@ -1,0 +1,90 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle.test;
+
+import java.io.File;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.openhab.tools.analysis.checkstyle.ImportExportedPackagesCheck;
+import org.openhab.tools.analysis.checkstyle.api.AbstractStaticCheckTest;
+
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.api.Configuration;
+import com.puppycrawl.tools.checkstyle.utils.CommonUtils;
+
+/**
+ * Tests for {@link ImportExportedPackagesCheck}
+ *
+ * @author Mihaela Memova
+ *
+ */
+public class ImportExportedPackagesCheckTest extends AbstractStaticCheckTest {
+
+    private static DefaultConfiguration configuration;
+
+    @BeforeClass
+    public static void setUp() {
+        configuration = createCheckConfig(ImportExportedPackagesCheck.class);
+    }
+
+    private static final String TEST_DIRECTORY_NAME = "importExportedPackagesCheckTest";
+    private static final String NOT_IMPORTED_PACKAGE_MESSAGE = "The exported package is not imported";
+
+    @Test
+    public void testManifestFileThatDoesNotImportAnExportedPackage() throws Exception {
+        String testFileName = "ManifestNotImportingAnExportedPackage.MF";
+        int lineNumber = 12;
+        String[] warningMessages = generateExpectedMessages(lineNumber, NOT_IMPORTED_PACKAGE_MESSAGE);
+        verifyManifestFile(testFileName, warningMessages);
+    }
+
+    @Test
+    public void testManifestFileThatDoesNotImportSeveralExportedPackages() throws Exception {
+
+        String testFileName = "ManifestNotImportingSeveralExportedPackages.MF";
+        int firstLineNumber = 14;
+        int secondLineNumber = 15;
+
+        String[] warningMessages = generateExpectedMessages(firstLineNumber, NOT_IMPORTED_PACKAGE_MESSAGE,
+                secondLineNumber, NOT_IMPORTED_PACKAGE_MESSAGE);
+        verifyManifestFile(testFileName, warningMessages);
+    }
+
+    @Test
+    public void testManifestFileThatDoesNotImportAnyPackages() throws Exception {
+
+        String testFileName = "ManifestNotImportingAnyPackages.MF";
+        int lineNumber = 10;
+
+        String[] warningMessages = generateExpectedMessages(lineNumber,
+                ImportExportedPackagesCheck.NOT_IMPORTED_PACKAGE_MESSAGE);
+        verifyManifestFile(testFileName, warningMessages);
+    }
+
+    @Test
+    public void testManifestThatImportAllExportedPackages() throws Exception {
+
+        String testFileName = "ManifestImportingAllExportedPackages.MF";
+        String[] warningMessages = CommonUtils.EMPTY_STRING_ARRAY;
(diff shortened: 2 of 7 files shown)


src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+		} catch (IOException e) {
+			e.printStackTrace();
+			log(0, String.format("An exception %s was thrown while processing the file: ", e, file.getPath()));
+		} catch (ParseException e) {
+			log(0, String.format("An exception %s was thrown while trying to parse the MANIFEST: ", e, file.getPath()));
+		}
+	}
+
+	private boolean setContainsObject(Set set, Object obj) {
+		for (Object o : set) {
```
martinvw: Does set have a contains method?

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+				if (!setContainsObject(imports, export)) {
+					int lineToLog = findLineNumber(lines, export.toString(), exportsStartIndex);
+					if(lineToLog != -1) {
+						log(lineToLog, NOT_IMPORTED_PACKAGE_MESSAGE);
+					}
+				}
+			}
+		} catch (IOException e) {
+			e.printStackTrace();
+			log(0, String.format("An exception %s was thrown while processing the file: ", e, file.getPath()));
```
martinvw: Only the e is used in string format...

And you now log it to the report, is that the correct place, or should we use a logger like in: https://github.com/openhab/static-code-analysis/blob/master/src/main/java/org/openhab/tools/analysis/checkstyle/MavenPomderivedInClasspathCheck.java

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+			for (Object export : exports) {
+				if (!setContainsObject(imports, export)) {
+					int lineToLog = findLineNumber(lines, export.toString(), exportsStartIndex);
+					if(lineToLog != -1) {
+						log(lineToLog, NOT_IMPORTED_PACKAGE_MESSAGE);
+					}
+				}
+			}
+		} catch (IOException e) {
+			e.printStackTrace();
```
martinvw: Print stacktrace should not be used.

src/test/java/org/openhab/tools/analysis/checkstyle/test/ImportExportedPackagesCheckTest.java
```diff
@@ -0,0 +1,81 @@
...
+                ImportExportedPackagesCheck.NOT_IMPORTED_PACKAGE_MESSAGE);
+        verifyManifestFile(testFileName, warningMessages);
+    }
+
+    public void verifyManifestFile(String fileName, String[] warningMessages) throws Exception {
+        String filePath = getPath(TEST_DIRECTORY_NAME + File.separator + fileName);
+        verify(configuration, filePath, warningMessages);
+    }
+
+    @Override
```
martinvw: There is no test for the happy scenario, so you did not prove that is able to give no errors at all ;-)

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+	protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
+		int exportsStartIndex = 0;
+		for (int i = 0; i < lines.size(); i++) {
+			if (lines.get(i).startsWith(EXPORT_PACKAGES_HEADER)) {
+				exportsStartIndex = i;
+			}
+		}
+
+		try {
+			Set exports = ManifestParser.parseManifest(file).getExports();
```
martinvw: According to the source I found both the set of Exports and the set of Imports should be typed.

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+			Set imports = ManifestParser.parseManifest(file).getImports();
+
+			for (Object export : exports) {
+				if (!setContainsObject(imports, export)) {
+					int lineToLog = findLineNumber(lines, export.toString(), exportsStartIndex);
+					if(lineToLog != -1) {
+						log(lineToLog, NOT_IMPORTED_PACKAGE_MESSAGE);
+					}
```
martinvw: Why not log it if -1 in what case will this happen and are you sure that you don't want to report anything then.

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+	public static final String EXPORT_PACKAGES_HEADER = "Export-Package:";
+
+	public ImportExportedPackagesCheck() {
+		setFileExtensions(MANIFEST_EXTENSTION);
+	}
+
+	@Override
+	protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
+		int exportsStartIndex = 0;
+		for (int i = 0; i < lines.size(); i++) {
```
martinvw: Why not use findLineNumber here, it does a contain, so I would suppose it suits the job? The only difference is that this one finds the last one instead of the first one.

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+			e.printStackTrace();
+			log(0, String.format("An exception %s was thrown while processing the file: ", e, file.getPath()));
+		} catch (ParseException e) {
+			log(0, String.format("An exception %s was thrown while trying to parse the MANIFEST: ", e, file.getPath()));
+		}
+	}
+
+	private boolean setContainsObject(Set set, Object obj) {
+		for (Object o : set) {
+			if ((o.toString()).equals(obj.toString())) {
```
martinvw: _Maybe you can add a small comment what you are actually doing here, I read the underlying source code and it makes complete sense what you do here, but the code first made me curious what was going on and why the normal Set.contains was not used :-)_

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+			}
+		} catch (IOException e) {
+			e.printStackTrace();
+			log(0, String.format("An exception %s was thrown while processing the file: ", e, file.getPath()));
+		} catch (ParseException e) {
+			log(0, String.format("An exception %s was thrown while trying to parse the MANIFEST: ", e, file.getPath()));
+		}
+	}
+
+	private boolean setContainsObject(Set set, Object obj) {
```
martinvw: Make typed and give proper names.

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+					if(lineToLog != -1) {
+						log(lineToLog, NOT_IMPORTED_PACKAGE_MESSAGE);
+					}
+				}
+			}
+		} catch (IOException e) {
+			e.printStackTrace();
+			log(0, String.format("An exception %s was thrown while processing the file: ", e, file.getPath()));
+		} catch (ParseException e) {
+			log(0, String.format("An exception %s was thrown while trying to parse the MANIFEST: ", e, file.getPath()));
```
martinvw: Same remark as line 60.

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+				exportsStartIndex = i;
+			}
+		}
+
+		try {
+			Set exports = ManifestParser.parseManifest(file).getExports();
+			Set imports = ManifestParser.parseManifest(file).getImports();
+
+			for (Object export : exports) {
+				if (!setContainsObject(imports, export)) {
```
martinvw: _A lot of people have problems see exclamations marks, most of time code reads better if you would just reverse the function of the method and rename it. if `importsDoesNotContain`_

m-memova: Thank you but I prefer to leave it that way. I found it more confusing to have a boolean method which returns true if something is not true (not there, has not and etc)

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
m-memova: In this case the logLine can never be -1. I check if some of the declared exports is not imported. So since the export is there, the method which searches its line in the file will always return a valid value.

martinvw: Than why the if-statement?

m-memova:
Thank you, @martinvw! I made some fixes corresponding to your comments. The PR is updated.

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,86 @@
...
+            if (lines.get(i).startsWith(EXPORT_PACKAGES_HEADER)) {
+                exportsStartIndex = i;
+            }
+        }
+
+        try {
+            Set<ExportPackage> exports = ManifestParser.parseManifest(file).getExports();
+            Set<BundleRequirement> imports = ManifestParser.parseManifest(file).getImports();
+
+            for (Object export : exports) {
```
martinvw: Iterate over type of `ExportPackage` here

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,86 @@
...
+            logger.error("An error occured while processing the file", e, file.getPath());
+        } catch (ParseException e) {
+            logger.error("An error occured while trying to parse the MANIFEST", e, file.getPath());
+        }
+    }
+
+    /**
+     * Checks if a particular package is imported. The default implementation of
+     * {@link Set#contains(Object)} cannot be used since the {@link Object#hashCode()}
+     * method is not overridden for the {@link BundleRequirement} class
```
martinvw: No this not the actual problem, there are two different classes at play, so they are not equal but refer to the same string value. That is why you should make everything typed. The `searchedPackage` is of type `ExportPackage` while using a list of type `Set<BundleRequirement> imports`

src/test/resources/checks/checkstyle/importExportedPackagesCheckTest/ManifestImportingAllExportedPackages.MF
```diff
@@ -0,0 +1,11 @@
...
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse SmartHome Example
+Bundle-SymbolicName: org.eclipse.smarthome.binding.example;singleton:=true
+Bundle-Vendor: Eclipse.org/SmartHome
+Bundle-Version: 0.9.0.qualifier
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Bundle-ClassPath: .
+Import-Package: org.openhab.first.package
+Service-Component: OSGI-INF/*.xml
+Export-Package: org.openhab.first.package
```
martinvw: Missing newline at the end of the file

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,86 @@
...
+    Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    public ImportExportedPackagesCheck() {
+        setFileExtensions(MANIFEST_EXTENSTION);
+    }
+
+    @Override
+    protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
+        int exportsStartIndex = 0;
+        for (int i = 0; i < lines.size(); i++) {
```
martinvw: Why not use `findLineNumber` here, it does a contain, so I would suppose it suits the job? The only difference is that this one finds the last one instead of the first one.

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
```diff
@@ -0,0 +1,86 @@
...
+            Set<BundleRequirement> imports = ManifestParser.parseManifest(file).getImports();
+
+            for (Object export : exports) {
+                if (!isPackageImported(imports, export)) {
+                    int lineToLog = findLineNumber(lines, export.toString(), exportsStartIndex);
+                    log(lineToLog, NOT_IMPORTED_PACKAGE_MESSAGE);
+                }
+            }
+        } catch (IOException e) {
+            logger.error("An error occured while processing the file", e, file.getPath());
```
martinvw: You use no placeholders and the `e` should be final argument, also apply to line 65

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
m-memova: You are right, I totally forgot about that :)

src/main/java/org/openhab/tools/analysis/checkstyle/ImportExportedPackagesCheck.java
m-memova: findLineNumber returns the line in the file (1-based) using the start line from the list (0-based). So if I use it, it should be : `exportsStartIndex = findLineNumber(lines, EXPORT_PACKAGES_HEADER, 0) - 1;` . And then a comment will be needed to explain why -1.

martinvw: Yeah, but what is the meaning of the field you call it an index but pass it to an argument which expects a line number. So I think that the following is way more correct and following the reader's expectations:

```java
int lineNumberBeforeExportsStart = findLineNumber(lines, EXPORT_PACKAGES_HEADER, 0) - 1;
```

src/test/java/org/openhab/tools/analysis/checkstyle/test/ImportExportedPackagesCheckTest.java
m-memova: I added a test case for that.

m-memova:
The PR is updated. I think all the @martinvw 's comments are addressed. As soon as this PR is merged I will create a PR for fixing the problems it found.

martinvw:
@m-memova thanks! LGTM

PR #82, opened 2017-06-02
fixed report location url

kaikreuzer:
The log returns an URI, which was missing the colon and which must always use forward slashes, independent of the OS.

Signed-off-by: Kai Kreuzer <kai@openhab.org>

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/report/ReportUtility.java
+++ b/src/main/java/org/openhab/tools/analysis/report/ReportUtility.java
@@ -286,8 +286,7 @@ public class ReportUtility extends AbstractMojo {
                 result.append(log);
             }
         }
-        result.append("Detailed report can be found at: file").append(File.separator).append(File.separator)
-                .append(File.separator).append(reportLocation).append("\n");
+        result.append("Detailed report can be found at: file:///").append(reportLocation).append("\n");
         return result.toString();
 
     }


martinvw:
Lgtm

src/main/java/org/openhab/tools/analysis/report/ReportUtility.java
```diff
@@ -286,8 +286,7 @@ private String checkForErrors(String secondMergeResult, String reportLocation) {
                 result.append(log);
             }
         }
-        result.append("Detailed report can be found at: file").append(File.separator).append(File.separator)
-                .append(File.separator).append(reportLocation).append("\n");
+        result.append("Detailed report can be found at: file:///").append(reportLocation).append("\n");
```
maggu2810: Relay three slashes aren't there two slashes between the scheme and the location and normally the first one if the location is an absolute one starting from the root (but part of the location)?

martinvw: Good question

kaikreuzer: For some reason the "reportLocation" is a path relative to the root, i.e. it does not start with a slash. So my PR just replaces three `File.separator`s by three forward slashes and adds the missing colon. The result is a valid URL that you can copy from the log and paste into the browser.
If you want to go into a deeper refactoring, feel free to do so as a follow-up PR.

martinvw: Thanks for the explanation 🙂

PR #87, opened 2017-06-08
Added exception for util classes in the exclude filter

### LINKED ISSUE #48: Static loggers in (util) classes
The build fails locally at (among other warnings (https://github.com/openhab/static-code-analysis/issues/47#issuecomment-291108909))
```
[ERROR] ERROR found by findbugs: org.openhab.binding.homematic.type.MetadataUtils.java:
[ERROR] Logger should be non-static field. Change this field (logger) to non-static field.
```
I think we should allow at least classes which are named util to have static loggers. 

Imho it has never been a bad thing to use static methods (even in normal classes) so there also some occurrences where we have other static methods which use logging. It feels strange and bad to refactor those static methods away because of this rule, can we make it a warning instead of an error.

svilenvul:
Closes #48

### DIFF
--- a/src/main/resources/rulesets/findbugs/exclude.xml
+++ b/src/main/resources/rulesets/findbugs/exclude.xml
@@ -15,4 +15,13 @@
   <Match>
     <Bug pattern="SLF4J_MANUALLY_PROVIDED_MESSAGE"/>
   </Match>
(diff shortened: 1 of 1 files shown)


martinvw:
Perfect, lgtm!

PR #110, opened 2017-07-13
Resource loading in MOJOs simplified.

svilenvul:
Remove the custom URLClassLoader and resource files loaded directly from the jar.


svilenvul:
Tested on Maven 3.0.5, no compatibility issues found.

If I recall correctly this was not working, when the project was part of the Maven build. This is not the case anymore.

PR #114, opened 2017-07-17
Replace findbugs deps with spotbugs

### LINKED ISSUE #111: FindBugs memory leak when same JVM is used
We have noticed that the static analysis tool is consuming a lot of memory when executed on a multi module build (as the openHAB build :)) and I made some research. I will try to summarize the results here.

I remember that we had a [discussion ](https://github.com/openhab/static-code-analysis/issues/15) if we should disable the fork option (which spawns a separate Java process for each execution of the FindBugs plugin), but this doesn't seem to be the best option now.

I have used the jconsole memory profiling tool and below can be seen the most critical leak that I have found (1 GB of non-heap memory usage):

sat-non-heap

The memory seems to be consumed by 100 000 classes that are being loaded (the lower left window). I suspect that the problem is related to the fact that the same JVM is used (see below for more details): 

sat

I have customized the build and excluding the FindBugs maven plugin the problem disappears. 

I have experimented again with a fork set to true and the results are good (non-heap memory usage is stable around 200 MB and 20 000 classes are loaded), the separate process consumes around 300 MB before dying:

satfork-non-heap
satfork

@kaikreuzer, I think the best option will be to use a separate thread for FindBugs, but this change could directly affect the openHAB build. Please share your opinion on the topic :)

### LINKED ISSUE #112: Switch from FindBugs to SpotBugs
Migrated from https://github.com/openhab/static-code-analysis/issues/15#issuecomment-315286317:

> kaikreuzer wrote :
>Btw (sorry, maybe slightly off-topic): Did you already consider switching to https://github.com/spotbugs/spotbugs?

[SpotBugs project description](https://spotbugs.github.io/):
> SpotBugs is the spiritual successor of FindBugs, carrying on from the point where it left off with support of its community.

[Requirements](https://spotbugs.readthedocs.io/en/latest/introduction.html#requirements):
>To use SpotBugs, you need a runtime environment compatible with Java version 1.8 or later. SpotBugs is platform independent, and is known to run on GNU/Linux, Windows, and MacOS X platforms.

>You should have at least 512 MB of memory to use SpotBugs. To analyze very large projects, more memory may be needed.

There is a [guide for migration](https://spotbugs.readthedocs.io/en/latest/migration.html#) from FindBugs to SpotBugs.

svilenvul:
Closes #112
Fixes #111

### DIFF
--- a/docs/maven-plugin.md
+++ b/docs/maven-plugin.md
@@ -106,7 +106,7 @@ Parameters:
 ### static-code-analysis:findbugs
 
 Description:
-    Executes the `findbugs-maven-plugin` goal `findbugs` with a  ruleset file and configuration properties
+    Executes the `spotbugs-maven-plugin` goal `findbugs` with a  ruleset file and configuration properties
 
 Parameters:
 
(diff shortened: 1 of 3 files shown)


docs/maven-plugin.md
```diff
@@ -106,7 +106,7 @@ Parameters:
 ### static-code-analysis:findbugs
 
 Description:
-    Executes the `findbugs-maven-plugin` goal `findbugs` with a  ruleset file and configuration properties
+    Executes the `spotbugs-maven-plugin` goal `findbugs` with a  ruleset file and configuration properties
```
martinvw: Same goal?

martinvw:
Looks great for the rest.

docs/maven-plugin.md
svilenvul: Yes, it sounds strange, but is the same [goal ](https://github.com/hazendaz/findbugs-maven-plugin/blob/master/src/main/groovy/org/codehaus/mojo/findbugs/FindBugsMojo.groovy#L58).

PR #123, opened 2017-08-01
Add exceptions for smarthome model bundles.

svilenvul:
Closes eclipse/smarthome#3926


kaikreuzer:
It is not really nice to have project specific configuration spoiling the general config here.
Wouldn't it be possible to add such exceptions somewhere in the ESH repo (once we add the analysis there)?

svilenvul:
I see, yes it is possible, but the both config files won't be merged (so some suppressions will be existing in both repos).

svilenvul:
Moving to the ESH repo.

PR #142, opened 2017-08-14
Update to Checkstyle 8.1

### LINKED ISSUE #140: Update Checkstyle to the newest version
Check the [change log](http://checkstyle.sourceforge.net/releasenotes.html) for breaking changes and update the used version from 7.7 to 8.1.

svilenvul:
Applied changes that are in sync with the Checkstyle API changes:
- processFiltered() accepts different arguments;
- added some unimplemented methods;

Closes #140

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/BuildPropertiesCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/BuildPropertiesCheck.java
@@ -24,6 +24,7 @@ import org.eclipse.pde.core.build.IBuildEntry;
 import org.openhab.tools.analysis.checkstyle.api.AbstractStaticCheck;
 
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+import com.puppycrawl.tools.checkstyle.api.FileText;
 
 import edu.emory.mathcs.backport.java.util.Collections;
 
@@ -96,18 +97,18 @@ public class BuildPropertiesCheck extends AbstractStaticCheck {
     }
 
     @Override
-    protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
+    protected void processFiltered(File file, FileText fileText) throws CheckstyleException {
         String fileName = file.getName();
         if (fileName.equals(BUILD_PROPERTIES_FILE_NAME)) {
             if (!isEmpty(file)) {
-                processBuildProperties(file, lines);
+                processBuildProperties(file, fileText.toLinesArray());
             } else {
                 log(0, EMPTY_FILE_MSG);
             }
         }
     }
 
-    private void processBuildProperties(File file, List<String> lines) throws CheckstyleException {
+    private void processBuildProperties(File file, String[] lines) throws CheckstyleException {
         // We ignore the exceptions thrown by the parseBuildProperties() method. A corrupt build.properties file will
         // fail the Maven build in the compile phase, so we should not care about this case in the validate phase
         IBuild buildPropertiesFile = parseBuildProperties(file);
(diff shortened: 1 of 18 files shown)


src/main/java/org/openhab/tools/analysis/checkstyle/DeclarativeServicesDependencyInjectionCheck.java
```diff
@@ -44,6 +44,11 @@
     }
 
     @Override
+    public int[] getAcceptableTokens() {
+        return getDefaultTokens();
```
dstivanov: Could it be that the default tokens are too much or too less for this check? Why not specifying only the needed ones?

src/main/java/org/openhab/tools/analysis/checkstyle/BuildPropertiesCheck.java
```diff
@@ -96,18 +97,18 @@ public void setPossibleSourceValues(String[] possibleSourceValues) {
     }
 
     @Override
-    protected void processFiltered(File file, List<String> lines) throws CheckstyleException {
+    protected void processFiltered(File file, FileText fileText) throws CheckstyleException {
         String fileName = file.getName();
         if (fileName.equals(BUILD_PROPERTIES_FILE_NAME)) {
             if (!isEmpty(file)) {
-                processBuildProperties(file, lines);
+                processBuildProperties(file, fileText.toLinesArray());
```
dstivanov: A bit sad that we have to move back to array when we have streams already :)

src/main/java/org/openhab/tools/analysis/tools/CheckstyleChecker.java
```diff
@@ -113,7 +113,7 @@ public void execute() throws MojoExecutionException {
         }
 
         // Maven may load an older version, if no version is specified
-        Dependency checkstyle = dependency("com.puppycrawl.tools", "checkstyle", "7.7");
+        Dependency checkstyle = dependency("com.puppycrawl.tools", "checkstyle", "8.1");
```
dstivanov: It would be great to have a way to get this version directly from the pom.xml, but as far as I know there is no other way to do it instead of reading the xml file directly...

src/main/java/org/openhab/tools/analysis/checkstyle/DeclarativeServicesDependencyInjectionCheck.java
svilenvul: As this check is too specific, I don't think that there are any cases when the acceptable tokens could be different that the default ones.

src/main/java/org/openhab/tools/analysis/checkstyle/BuildPropertiesCheck.java
svilenvul: Yes it is really sad, I couldn't find any use case for us for the newly added [FileText](http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/api/FileText.html) class.

src/main/java/org/openhab/tools/analysis/tools/CheckstyleChecker.java
svilenvul: Yes you are right, I couldn't find other solution. In this case maybe it is better to leave it as it is now.

src/main/java/org/openhab/tools/analysis/checkstyle/DeclarativeServicesDependencyInjectionCheck.java
dstivanov: OK for me, but I just wanted to be sure that this will not break if  the default tokens are changed in some of the newer versions.

src/main/java/org/openhab/tools/analysis/checkstyle/BuildPropertiesCheck.java
dstivanov: Probably the getCharset() will be convenient if we are reading free text (such as the Readme check)?

src/main/java/org/openhab/tools/analysis/tools/CheckstyleChecker.java
dstivanov: Yes, better leave it like this, but we should have this in mind.

PR #163, opened 2017-08-22
Small adjustments and bug fix to the external libraries checks.

VelinYordanov:


src/main/java/org/openhab/tools/analysis/checkstyle/OutsideOfLibExternalLibrariesCheck.java
```diff
@@ -26,6 +26,7 @@
  */
 public class OutsideOfLibExternalLibrariesCheck extends AbstractExternalLibrariesCheck {
     private static final String TARGET_FOLDER = "target";
+    private static final String BIN_FOLDER = "bin";
```
svilenvul: Could you add both as a value in a configuration property(`ignoredFolders`) ? Maybe we will have some other cases in the future and it will be great if we just can change the configuration.

PR #171, opened 2017-08-25
Ignored test correction

doandzhi:
Signed-off-by: Erdoan Hadzhiyusein <3rdoan@gmail.com>


martinvw:
The problem is in the capitals of the test file name README.**MD**' instead of 'README.md'

martinvw:
Linux has that problem MacOs and Windows are case insensitive on their file systems.

doandzhi:
Thanks a lot! 👍

svilenvul:
Thanks all.

PR #102, opened 2017-06-16
Added check for packages that are not exported and are not .internal

svilenvul:
Closes #100

Also-by: Aleksandar Kovachev <akovachev@tu-sofia.bg
Also-by: Petar Valchev <petar.valchev@musala.com>
Signed-off-by: Svilen Valkanov <svilen.valkanov@musala.com>

PR #172, opened 2017-08-29
Make check PackageExportsNameCheck work on none Windows system too.

Hilbrand:
backward slashes don't work well on none Windows systems.
Fix related to pr #102

Signed-off-by: Hilbrand Bouwkamp <hilbrand@h72.nl> (github: Hilbrand)

### DIFF
--- a/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/src/main/resources/rulesets/checkstyle/rules.xml
@@ -135,7 +135,7 @@
   
   <module name="org.openhab.tools.analysis.checkstyle.PackageExportsNameCheck">
     <property name="severity" value="warning" />
-    <property name="sourceDirectories" value="src\main\java" />
+    <property name="sourceDirectories" value="src/main/java" />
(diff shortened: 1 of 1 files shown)


svilenvul:
Thanks for this pull request. We have tests running on the Travis CI that use the same value and we haven't noticed it so far.

Could you give more information about the problem - stack trace or a link to a discussion where it is described ?

Hilbrand:
Actually the unit test doesn't use this value. It sets the value directly in the code and uses the same value as the value in my pull request and not as in the configuration file: https://github.com/openhab/static-code-analysis/blob/664495004e29f5c96e4e8adf41c7a5114c32d0fc/src/test/java/org/openhab/tools/analysis/checkstyle/test/PackageExportsNameCheckTest.java#L43

I ran into this problem while running it on an openHAB binding. I had modified my openHAB pom to use the 0.3.0-SNAPSHOT version and it didn't report any messages. Only after changing the value it did report them. I ran it on a Linux system.

PR #177, opened 2017-09-05
Fixed build problems.

VelinYordanov:


tools/suppressions.xml
```diff
@@ -18,4 +18,5 @@
     <suppress files=".+[\\/]gen[\\/].+\.java" checks="AuthorTagCheck"/>
     <!-- Some checks will be supressed for test bundles -->
     <suppress files=".+.test[\\/].+" checks="RequireBundleCheck"/>
+    <suppress files=".+org.openhab.tools.analysis.report.+|.+org.openhab.tools.analysis.findbugs.+" checks="RegexpHeaderCheck"/>
```
svilenvul: Please, be more precise with the exclusions (add the exact file name), we don`t want to skip the check for all classes in the package.

PR #182, opened 2017-09-08
Improvements for MarkdownCheck

svilenvul:
- dont read file twice - which caused character encoding missmatch


svilenvul:
@doandzhi, could you take a look ?

doandzhi:
That makes sense. As this is a usual practice in other checks as well maybe we would need some refactoring at some stage.

PR #189, opened 2017-09-15
Change the severity of PackageExportsNameCheck to error.

### LINKED ISSUE #100: PackageExportsNameCheck - severity error
@kaikreuzer 

We are writing a check that will cover this guideline:

> 10. Packages that contain classes that are not meant to be used by other bundles should have “internal” in their package name.  

What severity should be used ? Warning or maybe just info ?

Many bundles (according to my estimation around 40-50 percent of the bundles ) in the OH2 addons repo  contain packages that are not exported but are not "internal". Most often that are packages x.y.z.config and x.y.z.discovery. How important is this naming convention - should we proceed with some clean up and how ?

svilenvul:
Closes #100

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
@@ -136,7 +136,7 @@
   </module>
   
   <module name="org.openhab.tools.analysis.checkstyle.PackageExportsNameCheck">
-    <property name="severity" value="warning" />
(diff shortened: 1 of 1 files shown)


svilenvul:
I see no open Pull Requests linked to #100, so I guess so.

@VelinYordanov, are all issues that you have found in openHAB and ESH fixed ?

svilenvul:
I think it is the best to fix the remaining errors (if there are still any) before the new release of the tool.

PR #193, opened 2017-09-19
Generate warning when schedule at fixed rate is used

lpapazow:
Add supression for scheduleAtFixedRateCheck in wifiled bundle


martinvw:
Something went wrong with rebasing

PR #195, opened 2017-09-19
Initial contribution to the InheritDocCheck.

### LINKED ISSUE #103: Add a warning for JavaDocs that only use "inheritDoc"
I often see these in reviews:

```
   /**
    * {@inheritDoc}
    */
   @Override
```
e.g. [here](https://github.com/openhab/openhab2-addons/pull/2363/files#diff-7e55bfb77881a45bfe89e90c560e5187R137).

As Javadoc is inherited by default from an interface definition if nothing else is specified, this is redundant and should be avoided. I would therefore suggest to add a warning in such a case.

`inheritDoc` should imho only be used if some additional documentation is added to it on a specific method.

VelinYordanov:
Closes #103

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/InheritDocCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/InheritDocCheck.java
@@ -1,0 +1,59 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle;
+
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.CommonUtils;
+
+/**
+ * Check for unnecessary inheritDoc - As Javadoc is inherited by default from an interface definition if nothing else is
+ * specified, this is redundant and should be avoided. inheritDoc should only be used if some additional
+ * documentation is added to it on a specific method.
+ *
+ * @author Velin Yordanov - initial contribution
+ *
+ */
+public class InheritDocCheck extends AbstractCheck {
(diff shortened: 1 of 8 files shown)


src/main/java/org/openhab/tools/analysis/checkstyle/InheritDocCheck.java
```diff
@@ -0,0 +1,60 @@
...
+    public boolean isCommentNodesRequired() {
+        return true;
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        DetailAST javadocContentAST = ast.findFirstToken(TokenTypes.COMMENT_CONTENT);
+        String javadocContent = javadocContentAST.getText();
+        if (javadocContent.contains(INHERIT_DOC)) {
+            javadocContent = javadocContent.replaceAll(REGEX, "").trim();
```
martinvw: just use `replace("*", "")`

src/test/resources/checks/checkstyle/inheritDocCheckTest/fileWithALotOfNewLines.java
```diff
@@ -0,0 +1,15 @@
...
+         * 
+        * {@inheritDoc}
+        * 
+        * 
+        * 
+        */
+       public void shouldLogWhenThereIsAnEmptyInheritDoc() {
+           super.getPath("")
+       }
+   }
```
martinvw: Maybe add newlines in all the example files

VelinYordanov:
Thanks for the review. I updated the pull request.

src/main/java/org/openhab/tools/analysis/checkstyle/InheritDocCheck.java
```diff
@@ -0,0 +1,59 @@
...
+    public boolean isCommentNodesRequired() {
+        return true;
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        DetailAST javadocContentAST = ast.findFirstToken(TokenTypes.COMMENT_CONTENT);
+        String javadocContent = javadocContentAST.getText();
+        if (javadocContent.contains(INHERIT_DOC)) {
+            javadocContent = javadocContent.replaceAll("\\*", "").trim();
```
martinvw: I suggested `replace("*", "")` it takes just a plain string instead of regex and also replaces all occurrences, 

https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replace(java.lang.CharSequence,%20java.lang.CharSequence)

martinvw:
LGTM

Issue #197, opened 2017-09-20
Provide Bundle-Version Check


dstivanov:
As already identified in several places, some of the pull requests that are old are providind bundle version, that is out-of date. For example, the [Energenie Mi|Home binding PR](https://github.com/openhab/openhab2-addons/pull/2278) provides bundle version in the MANIFEST.MF
`Bundle-Version: 2.1.0.qualifier`
while the current release version is already
`Bundle-Version: 2.2.0.qualifier`

The same applies for the version in the pom.xml, which is usually defined as
`<version>2.1.0-SNAPSHOT</version>`

Can you please provide a check, that can be configured for a to check whether the Bundle-Version is specified correctly?

martinvw:
I think that check is there but since the SAT is always based on the code which you are running it has some problems in finding that out I would assume.

dstivanov:
@martinvw You are right - actually there is an option to include Regex for pom.xml and Bundle-version validation in the org.openhab.tools.analysis.checkstyle.PomXmlCheck. This will solve the problem if we set  the regex expressions for the PomXmlCheck.

I discussed with @svilenvul that these regex expressions are currently not set for the SAT nether for the pom.xml version, nor for the bundle-version and therefore cannot be overridden in openhab-addons or smarthome. Fortunately, we can fix it by introducing a general regex expression in the SAT and override it in the openhab-addons and smarthome repositories by altering the ruleset.properties files there.

VelinYordanov:
After this pull request the current version can be set with the manifestVersionRegularExpression property. However this means that if the version is changed the property must be changes as well.

svilenvul:
Already merged.

PR #204, opened 2017-10-10
Enhanced the PomXmlCheck. If provided a version regex it will now log…

VelinYordanov:
… when the current version is different.

See #197

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/PomXmlCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/PomXmlCheck.java
@@ -8,10 +8,7 @@
  */
 package org.openhab.tools.analysis.checkstyle;
 
-import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.MANIFEST_EXTENSION;
-import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.MANIFEST_FILE_NAME;
-import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.POM_XML_FILE_NAME;
-import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.XML_EXTENSION;
+import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.*;
 
 import java.io.File;
 import java.io.IOException;
@@ -42,6 +39,7 @@ import com.puppycrawl.tools.checkstyle.api.FileText;
  *
  * @author Petar Valchev - Initial implementation
  * @author Svilen Valkanov - Replaced headers, applied minor improvements, added check for parent pom ID
+ * @author Velin Yordanov - Enhanced the check. It can now check for a specific version.
  */
 public class PomXmlCheck extends AbstractStaticCheck {
     private static final String MISSING_VERSION_MSG = "Missing /project/version in the pom.xml file.";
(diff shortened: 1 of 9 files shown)


src/main/java/org/openhab/tools/analysis/checkstyle/PomXmlCheck.java
```diff
@@ -137,6 +134,10 @@ private void processManifestFile(FileText fileText) throws CheckstyleException {
         // We need this in order to filter the "qualifier" for the Snapshot versions
         manifestVersion = getVersion(version.toString(), manifestVersionPattern);
 
+        if (manifestVersion == null) {
+            log(0, "The manifest version did not match the requirements." + manifestVersionRegularExpression);
```
svilenvul: Could you add the message as constant above.

svilenvul:
I see you have added a test case, where both the pom.xml version and the MANIFEST.MF version do not match, but what about if one of both is wrong ? Do you think it makes sense to cover it with two additional test cases ?

VelinYordanov:
@svilenvul If only 1 is wrong the check will log an error because the versions are not the same and the check already looks for that.

src/test/java/org/openhab/tools/analysis/checkstyle/test/PomXmlCheckTest.java
```diff
@@ -98,18 +99,31 @@ public void testMasterPom() throws Exception {
...
+
     @Override
     protected DefaultConfiguration createCheckerConfig(Configuration config) {
         DefaultConfiguration defaultConfiguration = new DefaultConfiguration("root");
         defaultConfiguration.addChild(config);
         return defaultConfiguration;
     }
 
-    private void verifyPomXmlFile(String testDirectoryName, int expectedLine, String expectedMessage) throws Exception {
+    private void verifyFile(String testDirectoryName, boolean isPomXmlFile, int expectedLine, String expectedMessage)
```
svilenvul: Adding a boolean argument makes these tests harder to read and understand.

Could you just add a separate method `verifyManifestFile` ?

src/test/java/org/openhab/tools/analysis/checkstyle/test/PomXmlCheckTest.java
```diff
@@ -98,18 +99,31 @@ public void testMasterPom() throws Exception {
         verify(createChecker(config), testFileAbsolutePath, ArrayUtils.EMPTY_STRING_ARRAY);
     }
 
+    @Test
+    public void testDifferentVersion() throws Exception {
+        verifyFile("different_version_directory", false, 0,
+                "The manifest version did not match the requirements " + MANIFEST_REGEX);
```
svilenvul: Could you add it to the other messages above as constant?

src/test/java/org/openhab/tools/analysis/checkstyle/test/PomXmlCheckTest.java
```diff
@@ -98,18 +99,31 @@ public void testMasterPom() throws Exception {
...
+            throws Exception {
         String testDirectoryAbsolutePath = getPath(POM_XML_CHECK_TEST_DIRECTORY_NAME + "/" + testDirectoryName);
         File testDirectory = new File(testDirectoryAbsolutePath);
         File[] testFiles = listFilesForDirectory(testDirectory, new ArrayList<File>());
-        String testFilePath = testDirectory.getPath() + File.separator + POM_XML_FILE_NAME;
+        String testFilePath = null;
+        if (isPomXmlFile) {
+            testFilePath = testDirectory.getPath() + File.separator + POM_XML_FILE_NAME;
+        } else {
+            testFilePath = testDirectory.getPath() + File.separator + "META-INF" + File.separator + "MANIFEST.MF";
```
svilenvul: You can use the constants placed in `org.openhab.tools.analysis.checkstyle.api.CheckConstants`.

src/main/java/org/openhab/tools/analysis/checkstyle/PomXmlCheck.java
```diff
@@ -137,6 +135,10 @@ private void processManifestFile(FileText fileText) throws CheckstyleException {
         // We need this in order to filter the "qualifier" for the Snapshot versions
         manifestVersion = getVersion(version.toString(), manifestVersionPattern);
 
+        if (manifestVersion == null) {
+            log(0, String.format(INCORRECT_VERSION, manifestVersionRegularExpression));
```
svilenvul: OK, but what if the manifest version is correct and the pom version is not ? I would expect a similar message (e.g. `The pom.xml version did not match the requirements`).

Please add both test cases:
- pom.xml version does not match, but manifest version matches
- manifest version does not match, but pom version matches

Then it will be visible what message is logged in each case.

PR #207, opened 2017-10-12
Fixed a problem where the ParameterizedRegexpHeaderCheck would not di…

VelinYordanov:
…splay the pattern in the error message properly.

It used to display the error message without properly formatting the pattern like this:
`Header line doesnt match pattern {0}`.

With this change it formats it correctly.
You can see a familiar issue [here](https://stackoverflow.com/questions/17569608/format-a-message-using-messageformat-format-in-java)


svilenvul:
Great, thanks !

Issue #205, opened 2017-10-11
A list ending cannot be processed properly


martinvw:
When running a check on https://github.com/openhab/openhab2-addons/tree/master/addons/binding/org.openhab.binding.rfxcom

I get

> [ERROR] A list ending cannot be processed properly: HQ_COCO20 - HQ COCO-20
[ERROR] A list ending cannot be processed properly: RGB_TRC02_2 - RGB TRC02_2 (3 batt)
[ERROR] A list ending cannot be processed properly: RFY_EXT - RFY Ext
[ERROR] A list ending cannot be processed properly: DIGIMAX_SHORT - Digimax with short format (no set point)
[ERROR] A list ending cannot be processed properly: HOME_CONFORT - Home Confort

But to me it does not look that fishy, wdyt?

```
#### Configuration Options

 * deviceId - Device Id
    * Device Id. House code + unit code, separated by dot. Example A.1

 * subType - Sub Type
    * Specifies device sub type.

        * X10 - X10 lighting
        * ARC - ARC
        * AB400D - ELRO AB400D (Flamingo)
        * WAVEMAN - Waveman
        * EMW200 - Chacon EMW200
        * IMPULS - IMPULS
        * RISINGSUN - RisingSun
        * PHILIPS - Philips SBC
        * ENERGENIE - Energenie ENER010
        * ENERGENIE\_5 - Energenie 5-gang
        * COCO - COCO GDR2-2000R
        * HQ\_COCO20 - HQ COCO-20


### lighting2 - RFXCOM Lighting2 Actuator

A Lighting2 device.
```

PR #209, opened 2017-10-25
fix escaped symbols breaking markdown checks

lpapazow:
Create a fix for [issue #205 ](https://github.com/openhab/static-code-analysis/issues/205).

Fix a single backslash before special symbols (!"#$%&\'()*+,./:;<=>?@][\^_`{|}~-) not escaping that symbol in .markdown files.

### DIFF
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
@@ -40,6 +40,7 @@ import org.eclipse.jface.text.IDocument;
 import org.eclipse.pde.core.build.IBuild;
 import org.eclipse.pde.internal.core.text.build.BuildModel;
 import org.jsoup.Jsoup;
+import org.openhab.tools.analysis.utils.LineFormatterFunction;
 import org.w3c.dom.Document;
 import org.xml.sax.SAXException;
 
@@ -54,10 +55,44 @@ import com.puppycrawl.tools.checkstyle.api.MessageDispatcher;
  * @author Svilen Valkanov - Initial contribution, add Exception to findLineNumber method
  * @author Mihaela Memova - Simplify findLineNumber method
  * @author Velin Yordanov - Used FileText instead of File to avoid unnecessary IO
+ * @author Lyubomir Papazov - Add method findLineNumberFormatted that takes a callback function that formats each line
  */
 public abstract class AbstractStaticCheck extends AbstractFileSetCheck {
 
     private Log logger = LogFactory.getLog(AbstractStaticCheck.class);
+
+    /**
+     * Finds the first occurrence of a text in a list of text lines representing the file content and
+     * returns the line number, where the text was found
+     *
+     *
+     * @param fileContent - represents the text content
+     * @param searchedText - the text that we are looking for
+     * @param startLineNumber - the line number from which the search starts exclusive, to start the
+     *            search of the beginning of the text the startLineNumber should be 0
+     * @param lineFormatterFunction - a callback function to format the line before trying to find the searched text
+     * @return the number of the line starting from 1, where the searched text occurred for the first
+     *         time
+     * @throws NoResultException when no match was found
+     */
+    protected int findLineNumberFormatted(FileText fileContent, String searchedText, int startLineNumber,
+            LineFormatterFunction lineFormatterFunction) throws NoResultException {
+        for (int lineNumber = startLineNumber; lineNumber < fileContent.size(); lineNumber++) {
+            String line = fileContent.get(lineNumber);
+
+            line = lineFormatterFunction.formatLine(line);
+
+            if (line.contains(searchedText)) {
+                // The +1 is to compensate the 0-based list and the 1-based text file
+                return lineNumber + 1;
+            }
+        }
+        String message = MessageFormat.format(
+                "`{0}` was not found in the file {1} starting from line `{2}`."
+                        + " Check if it is split between multiple lines or it is missing",
+                searchedText, fileContent.getFile().getAbsolutePath(), startLineNumber);
+        throw new NoResultException(message);
+    }
 
     /**
      * Finds the first occurrence of a text in a list of text lines representing the file content and
@@ -74,18 +109,7 @@ public abstract class AbstractStaticCheck extends AbstractFileSetCheck {
      */
     protected int findLineNumber(FileText fileContent, String searchedText, int startLineNumber)
             throws NoResultException {
-        for (int lineNumber = startLineNumber; lineNumber < fileContent.size(); lineNumber++) {
-            String line = fileContent.get(lineNumber);
-            if (line.contains(searchedText)) {
-                // The +1 is to compensate the 0-based list and the 1-based text file
-                return lineNumber + 1;
-            }
-        }
-        String message = MessageFormat.format(
-                "`{0}` was not found in the file {1} starting from line `{2}`."
-                        + " Check if it is split between multiple lines or it is missing",
-                searchedText, fileContent.getFile().getAbsolutePath(), startLineNumber);
-        throw new NoResultException(message);
+        return findLineNumberFormatted(fileContent, searchedText, startLineNumber, s -> s);
     }
 
     /**
--- a/src/main/java/org/openhab/tools/analysis/checkstyle/readme/MarkdownCheck.java
+++ b/src/main/java/org/openhab/tools/analysis/checkstyle/readme/MarkdownCheck.java
@@ -8,17 +8,15 @@
  */
 package org.openhab.tools.analysis.checkstyle.readme;
 
-import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.BIN_INCLUDES_PROPERTY_NAME;
-import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.BUILD_PROPERTIES_FILE_NAME;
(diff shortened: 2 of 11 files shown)


src/main/java/org/openhab/tools/analysis/utils/LineFormatterCallback.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * A callback interface for {@link MarkdownCheck}
+ *
+ * @author Lyubomir Papazov - initial contribution
+ *
+ */
+public interface LineFormatterCallback {
+    /**
+     * Used in AbstractStaticCheck's findLineNumber method to apply rules
+     * about escaping special characters to a line in the .mardown file
```
svilenvul: typo .mardown -> markdown 

But maybe it is better .md as this is the file extension that we are using in the check, or just say from the markdown file (without the .)

src/main/java/org/openhab/tools/analysis/utils/LineFormatterCallback.java
```diff
@@ -0,0 +1,28 @@
...
+     *
+     * @param line - an unparsed line from the .markdown file
```
svilenvul: Same as above

src/main/java/org/openhab/tools/analysis/checkstyle/readme/MarkdownCheck.java
```diff
@@ -94,23 +93,54 @@ private void checkReadMe(FileText fileText) {
...
             public void log(int line, String message) {
                 MarkdownCheck.this.log(line, message);
             }
         };
-        MarkdownVisitor visitor = new MarkdownVisitor(callBack, fileText);
+        MarkdownVisitor visitor = new MarkdownVisitor(callBack, fileText, new LineFormatterCallback() {
+
+            @Override
+            public String formatLine(String line) {
+                Pattern needsEscape = Pattern.compile("\\\\" + Escaping.ESCAPABLE);
```
martinvw: But you can not be sure that all characters which have to escaped are actually escaped are you sure this will suffice for the most common or better all cases? Should we maybe determine the line number in a different way instead of trying to match it again?

martinvw: Given such changes I would expect new tests? Or maybe changed tests?

src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
```diff
@@ -68,14 +70,20 @@
...
      * @param startLineNumber - the line number from which the search starts exclusive, to start the
      *            search of the beginning of the text the startLineNumber should be 0
+     * @param lineFormatterCallback - a callback function to format the line before trying to find the searched text
      * @return the number of the line starting from 1, where the searched text occurred for the first
      *         time
      * @throws NoResultException when no match was found
      */
-    protected int findLineNumber(FileText fileContent, String searchedText, int startLineNumber)
-            throws NoResultException {
+    protected int findLineNumberFormatted(FileText fileContent, String searchedText, int startLineNumber,
```
martinvw: Imho this is not the best way forward it looks a little over designed to me with this callback logic, isn't there an easy way to solve this?

src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
```diff
@@ -92,6 +100,24 @@ protected int findLineNumber(FileText fileContent, String searchedText, int star
...
+     * @param searchedText - the text that we are looking for
+     * @param startLineNumber - the line number from which the search starts exclusive, to start the
+     *            search of the beginning of the text the startLineNumber should be 0
+     * @return the number of the line starting from 1, where the searched text occurred for the first
+     *         time
+     * @throws NoResultException when no match was found
+     */
+    protected int findLineNumber(FileText fileContent, String searchedText, int startLineNumber)
+            throws NoResultException {
+        return findLineNumberFormatted(fileContent, searchedText, startLineNumber, null);
```
martinvw: I don't like `null` I would have prefered a no-op `LineFormatterCallback` if we go for this solution than you also do not need the null check, just let it explode if you are called with a `null`

src/main/java/org/openhab/tools/analysis/checkstyle/readme/MarkdownCheck.java
```diff
@@ -94,23 +93,54 @@ private void checkReadMe(FileText fileText) {
...
+                        lineFormatterCallback);
             }
 
             @Override
             public void log(int line, String message) {
                 MarkdownCheck.this.log(line, message);
             }
         };
-        MarkdownVisitor visitor = new MarkdownVisitor(callBack, fileText);
+        MarkdownVisitor visitor = new MarkdownVisitor(callBack, fileText, new LineFormatterCallback() {
```
martinvw: If the design stays as is, I would make the `LineFormatterCallback` a [`FunctionalInterface`](http://www.baeldung.com/java-8-functional-interfaces)

src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
martinvw: It also seems that source from which we receive the string has not the same origin as lines we are searching through which seems dangerous for a lot of reasons imho.

src/main/java/org/openhab/tools/analysis/checkstyle/readme/MarkdownCheck.java
lpapazow: The **Escaping.ESCAPABLE** constant is defined in the external markdown library and is used in the markdown parsing logic.  It will cover all the cases in which the markdown parser will make changes to the original text. 

The reason why I didn't include any new tests is that when I use the **verify** method that is provided by **com.puppycrawl.tools.checkstyle.BaseCheckTestSupport** and I provide no error messages to that method, it doesn't fail the test, it only logs the error. Thus, every test that I include, would pass even if it shouldn't. 

Do you have any ideas how to create tests that could fail?

src/main/java/org/openhab/tools/analysis/checkstyle/readme/MarkdownCheck.java
lpapazow: Great suggestion, I will use it.

src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
lpapazow: Thank you for this suggestion, it does seem to make the code more robust. I will rewrite it in the suggested way.

src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
lpapazow: The sources differ is because the **searchedText** is generated by the markdown parsing logic and the **fileContent** is the raw text in the markdown file.

martinvw: Yes, I did understand that, but isn't this part of the problem and the thing we would actually like to resolve, can't we just request the parsed text from markdown and search through that?>

src/main/java/org/openhab/tools/analysis/checkstyle/readme/MarkdownCheck.java
martinvw: Can you please discuss it with some of your college's first

src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheck.java
dstivanov: @martinvw I will try to jump in here, because we discussed it in the office with @lpapazow and @doandzhi : Originally the check is reporting errors into the raw markdown file and for this reason we want to find the exact line in the raw file and report the error there. So if we get the parsed text and report the errors into its lines, we will probably hit an inconsistent line. For this reason we decided to parse a bit the special symbols (taken from the parser) in order to find the correct raw content. What do you think, are you tend to agree or disagree ?

martinvw: I understand what we try to achieve but it feels a little fragile to me, in most languages there are more character which can be escaped than the ones you have to escape so mismatches are possible. Without any proper tests that is hard to tell without me diving into the code myself. 

I found an issue referring to the actual underlying problem that the markdown `Node`s do not have a reference back to the original location: https://github.com/atlassian/commonmark-java/issues/1

Given that the issue is there, seems it cannot be done (easily) so we might go for this current solution but not without proper tests.

martinvw: Note that good example where escaping is not always performed can be seen here:

http://docs.openhab.org/addons/bindings/netatmo/readme.html
https://github.com/openhab/openhab2-addons/edit/master/addons/binding/org.openhab.binding.netatmo/README.md on line 49 there is a unescaped _ which does not cause problems in some views but in others it does. Docs renders fine, github view renders fine but diff's don't.

svilenvul:
@lpapazow any updates on this PR ?

svilenvul:
Replaced by https://github.com/openhab/static-code-analysis/pull/245

PR #212, opened 2017-11-02
Add check to generate warning if ThingHandler sets Thing label

### LINKED ISSUE #198: Provide a Check for Thing Handler Trying to Set the Thing Label
This could appear pretty specific check for Eclipse SmartHome, but as it could be a common mistake I want to propose it. 

Currently the org.eclipse.smarthome.core.thing.binding.BaseThingHandler provide a protected field of type org.eclipse.smarthome.core.thing.Thing, which can be used be used to set the label of the thing programatically. However, the label have to be set only by the user and the binding must not overwrite that.

Therefore I will request a check that identifies such situations in the Thing handlers code.

tanyatanyatanya:
Add new check to generate a warning if a class that extends BaseThingHandler tries to set the Thing label.
Fixed #198

### DIFF
--- a/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
+++ b/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
@@ -1,0 +1,212 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle;
+
+import static org.openhab.tools.analysis.checkstyle.api.CheckConstants.JAVA_EXTENSION;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Optional;
+import java.util.Queue;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openhab.tools.analysis.checkstyle.api.AbstractStaticCheck;
+
+import com.puppycrawl.tools.checkstyle.TreeWalker;
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.FileContents;
+import com.puppycrawl.tools.checkstyle.api.FileText;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.CheckUtils;
+
+import antlr.RecognitionException;
+import antlr.TokenStreamException;
+
+/**
+ * Checks the code and generates a warning if a class extends BaseThingHandler
+ * and the instance method setLabel of the Thing interface is used.
+ * The label have to be set only by the user.
+ *
+ * @author Tanya Georgieva - Initial Contribution
+ */
+public class ThingHandlerSetLabelCheck extends AbstractStaticCheck {
+
+    private static final String WARNING_MESSAGE = "Do not use the setLabel of the thing in the ThingHandler";
+    private static final String BASE_THING_HANDLER = "BaseThingHandler";
+    private static final String SEARCHED_PACKAGE = "org.eclipse.smarthome.core.thing.binding." + BASE_THING_HANDLER;
+    private static final String SET_LABEL = "setLabel";
+    private static final String THING = "thing";
+    private static final String UNKNOWN = "UNKNOWN";
+
+    private final Log logger = LogFactory.getLog(ThingHandlerSetLabelCheck.class);
+
+    private String fullJavaExtension = "\\." + JAVA_EXTENSION;
+
+    private HashMap<File, DetailAST> files;
+    private HashMap<File, String> derivedAndBaseClasses;
+    private HashMap<File, DetailAST> classesExtendingBaseThingHandler;
+
+    public ThingHandlerSetLabelCheck() {
+        setFileExtensions(JAVA_EXTENSION);
+    }
+
+    @Override
+    public void init() {
+        files = new HashMap<>();
+        derivedAndBaseClasses = new HashMap<>();
+        classesExtendingBaseThingHandler = new HashMap<>();
+    }
+
+    @Override
+    protected void processFiltered(File file, FileText fileText) throws CheckstyleException {
+        FileContents fileContents = new FileContents(fileText);
+        DetailAST rootAST = null;
+
+        try {
+            rootAST = TreeWalker.parse(fileContents);
+        } catch (RecognitionException | TokenStreamException e) {
+            logger.error("Parsing file not successful");
+            e.printStackTrace();
+        }
+        files.put(file, rootAST);
+    }
+
+    @Override
+    public void finishProcessing() {
+        findDerivedAndBaseClasses();
+        findClassesExtendingBaseThingHandler();
+
+        classesExtendingBaseThingHandler.forEach((file, rootAST) -> {
+            String fileName = file.getName();
+            Path filePath = file.toPath();
+            checkSetLabelMethodCall(rootAST, fileName, filePath);
+        });
+    }
+
+    // Pre-order traversal - starting from the root first
+    private void checkSetLabelMethodCall(DetailAST nodeAST, String fileName, Path filePath) {
+        if (nodeAST == null) {
+            return;
+        }
+
+        if (THING.equals(nodeAST.getText())) {
+            // thing.setLabel(...)
+            boolean isThingFieldSetLabelMethodCalled = nodeAST.getNextSibling() != null
+                    && SET_LABEL.equals(nodeAST.getNextSibling().getText());
+            // when literal THIS is used the setLabel method is sibling of the parent of the node - thing
+            // this.thing.setLabel(...)
+            boolean isReferredThingFieldSetLabelMethodCalled = nodeAST.getParent().getNextSibling() != null
+                    && SET_LABEL.equals(nodeAST.getParent().getNextSibling().getText());
+            boolean isSetLabelCalled = isThingFieldSetLabelMethodCalled || isReferredThingFieldSetLabelMethodCalled;
+
+            if (isSetLabelCalled) {
+                int nodeLineNumber = nodeAST.getLineNo();
+                String fileStringPath = filePath.toString();
+                logMessage(fileStringPath, nodeLineNumber, fileName, WARNING_MESSAGE);
+            }
+        }
+        checkSetLabelMethodCall(nodeAST.getFirstChild(), fileName, filePath);
+        checkSetLabelMethodCall(nodeAST.getNextSibling(), fileName, filePath);
+    }
+
+    public void findClassesExtendingBaseThingHandler() {
+        Queue<File> visitedFiles = new LinkedList<>();
+        File searchedPackageFile = new File(SEARCHED_PACKAGE);
+        visitedFiles.add(searchedPackageFile);
+
+        while (!visitedFiles.isEmpty()) {
+            File currentFile = visitedFiles.poll();
+            String currentFileParent = currentFile.getParent();
+            String currentFileClassName = currentFile.getName().replaceFirst(fullJavaExtension, "");
+
+            derivedAndBaseClasses.forEach((file, baseClassName) -> {
+                String fileParent = file.getParent();
+                // Check if the current class and it's base class are from the same package.
+                // To find first the classes extending BaseThingHandler directly,
+                // we set the first file to be the SEARCHED PACKAGE,
+                // it's parent is null so the check is for the currentFileClassName
+                boolean areFilesFromSamePackage = SEARCHED_PACKAGE.equals(currentFileClassName)
+                        || currentFileParent.equals(fileParent);
+
+                if (currentFileClassName.equals(baseClassName) && areFilesFromSamePackage) {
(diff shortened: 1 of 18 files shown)


src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
```diff
@@ -0,0 +1,134 @@
...
+        return strBuilder.toString();
+    }
+
+    private boolean isHandlerExtendingBaseThingHandler(DetailAST ast) {
+        DetailAST extendsClause = ast.findFirstToken(TokenTypes.EXTENDS_CLAUSE);
+
+        if (extendsClause != null) {
+            String inheritedClassName = extendsClause.getFirstChild().getText();
+            // Only the simple name of the class can be retrieved from the extends clause
+            if (inheritedClassName.equals(BASE_THING_HANDLER)) {
```
martinvw: Does this work if you extend from the class indirectly? Also the import check looks non-robust to that. WDYT?

tanyatanyatanya: That's a great idea. I will check this case.

kaikreuzer:
@martinvw Does this look good to you now?

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
```diff
@@ -0,0 +1,212 @@
...
+            DetailAST currentNode = visitedNodes.poll();
+
+            if (currentNode.getType() == TokenTypes.CLASS_DEF) {
+                DetailAST extendsClause = currentNode.findFirstToken(TokenTypes.EXTENDS_CLAUSE);
+
+                if (extendsClause != null) {
+                    String inheritedClassName = extendsClause.getFirstChild().getText();
+                    // Only the simple name of the class can be retrieved from the extends clause
+                    return Optional.of(inheritedClassName);
+                }
```
martinvw: Should you maybe return here as well? Or should we check for all classes in a single file what if we we put the parent and child class in a single file and one extra to make it hard, will it work?

```java
class Test extends Object {
}

public class SomeThingHandler extends TestThingHandler { 
   public SomeThingHandler(){
      thing.setLabel("test");
   }
}

public class TestThingHandler extends BaseThingHandler {
}

File: custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java [hunk 3]
```diff
@@ -0,0 +1,212 @@
...
+import antlr.TokenStreamException;
+
+/**
+ * Checks the code and generates a warning if a class extends BaseThingHandler
+ * and the instance method setLabel of the Thing interface is used.
+ * The label have to be set only by the user.
+ *
+ * @author Tanya Georgieva - Initial Contribution
+ */
+public class ThingHandlerSetLabelCheck extends AbstractStaticCheck {
```
martinvw: I don't know about the performance of this but I personally doubt whether this check (with some corner cases which are extremely hard to cover) is worth iterating every file tree at least 1 additional time and every handler at least one additional time extra, @dstivanov wdyt?

@tanyatanyatanya Can you give some idea of the build time this would add to for example openHAB or ESH?

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private static final String WARNING_MESSAGE = "Do not use the setLabel of the thing in the ThingHandler";
+    private static final String BASE_THING_HANDLER = "BaseThingHandler";
+    private static final String SEARCHED_PACKAGE = "org.eclipse.smarthome.core.thing.binding." + BASE_THING_HANDLER;
```
martinvw: This is not a package this is a fully qualified name

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    private void findDerivedAndBaseClasses() {
+        files.forEach((file, rootAST) -> {
+            Optional<String> fullyQualifiedName = getFullyQualifiedName(rootAST);
+            Optional<String> baseClassName = getBaseClassName(rootAST);
+            String currentFullName = fullyQualifiedName.orElse(UNKNOWN);
+            String currentBaseName = baseClassName.orElse(UNKNOWN);
+            // if the currentFullName is not UNKNOWN our SEARCHED PACKAGE is found for the current class
+            boolean isHandlerImportsBaseThingHanlderPackage = !UNKNOWN.equals(currentFullName);
```
martinvw: Typo 'Hanlder'

custom-checks/checkstyle/src/test/resources/checkstyle/thingHandlerSetLabelCheckTest/org.openhab.binding.extending.handler/ThingHandlerExtendingIncorrectClass.java
```diff
@@ -0,0 +1,18 @@
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.smarthome.core.thing.ChannelUID;
+import org.eclipse.smarthome.core.thing.Thing;
+import org.eclipse.smarthome.core.thing.binding.BaseThingHandler;
+import org.example.test.core.thing.binding.ExampleAbstractClass;
+import org.eclipse.smarthome.core.types.Command;
+
+public class ThingHandlerExtendingIncorrectClass extends ExampleAbstractClass{
```
martinvw: Please format all test file I especially noted the missing spaces in front of the {

custom-checks/checkstyle/src/test/resources/checkstyle/thingHandlerSetLabelCheckTest/org.openhab.binding.extending.handler/ThingHandlerExtendingIndirectlyIncorrectBTH.java
```diff
@@ -0,0 +1,16 @@
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.smarthome.core.thing.ChannelUID;
+import org.eclipse.smarthome.core.thing.Thing;
+import org.eclipse.smarthome.core.types.Command;
+
+public class ThingHandlerExtendingIndirectlyIncorrectBTH extends ThingHandlerWithSetLabel{
```
martinvw: Will this file get a warning, would it also get a warning if it extended `ThingHandlerWithThingFieldWithoutSetLabel`?

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
```diff
@@ -0,0 +1,212 @@
...
+
+    // Pre-order traversal - starting from the root first
+    private void checkSetLabelMethodCall(DetailAST nodeAST, String fileName, Path filePath) {
+        if (nodeAST == null) {
+            return;
+        }
+
+        if (THING.equals(nodeAST.getText())) {
+            // thing.setLabel(...)
+            boolean isThingFieldSetLabelMethodCalled = nodeAST.getNextSibling() != null
```
martinvw: If we have a custom thing which we call thing and call `setLabel` on that will that be allowed or report a false positive?

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
```diff
@@ -0,0 +1,212 @@
...
+        } catch (RecognitionException | TokenStreamException e) {
+            logger.error("Parsing file not successful");
+            e.printStackTrace();
+        }
+        files.put(file, rootAST);
+    }
+
+    @Override
+    public void finishProcessing() {
+        findDerivedAndBaseClasses();
```
martinvw: I think / suppose that extension could easily bypass the boundaries of maven projects and AFAIK this test is executed on maven module level. So if the parent class (B) of our subject (A) which does extend the target class (C) lives in another module it will fail. 
Module 1:
A
Module 2:
B
Module 3 (ESH):
C

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
dstivanov: @martinvw I discussed with Tanya that we will provide measurements for the build times without this check, with the initial changes and with the new ones (which requires more files processing). When this is in place we will have an objective argument which way to go. I still believe that the current version will not be so time-consuming as it is working on bundle level and filters out the files. Also, the increased build times recently are subject of our investigations currently (see here - https://github.com/openhab/static-code-analysis/issues/232) so I guess that soon they will be improved.

martinvw: But my main problem is that imho its not able (not even possible?) to cover all scenarios so that it even takes a limited time without providing real safety might be too much already.

kaikreuzer:
FTR, I suggest I do the new SAT release without this feature, so that there is enough time for the further discussion here.

custom-checks/checkstyle/src/test/resources/checkstyle/thingHandlerSetLabelCheckTest/org.openhab.binding.extending.handler/ThingHandlerExtendingIndirectlyIncorrectBTH.java
tanyatanyatanya: No, it will not get a warning. For the first case the check will not even parse it because the ThingHandlerWithSetLabel class here extends incorrect BTH. And for the second case the check will create a new HashMap of files for the current directory: in this test case "org.openhab.binding.extending.handler" and ThingHandlerWithThingFieldWithoutSetLabel is from "org.openhab.binding.thing.handler"

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
tanyatanyatanya: No, it will not work because I am currently searching for the first extend.

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
tanyatanyatanya: As I start searching first for the correct BTH in the classes that extends it directly (after that I am searching for the classes with indirect inheritance) if such classes are found I continue with the thing and setLabel method search. And again if any,  I rely on that there won't be any other instance called exactly "thing" or method "setLabel" in the current class. But yes, if there is such custom thing with the same method name it will be reported.

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ThingHandlerSetLabelCheck.java
tanyatanyatanya: The build time without this check on openHAB is 17:09 minutes, with this new version of the check is 20:55 minutes and with the old version where I do not search for indirect inheritance is 18:45 minutes.

svilenvul:
I feel so bad that this PR is covered with dust here, it would be great if we can make some progress.

@tanyatanyatanya has provided measurements for the build times without this check, with the initial changes and with the new ones as @dstivanov  proposed :

```
tanyatanyatanya on Jan 10

The build time without this check on openHAB is 17:09 minutes,
with this new version of the check is 20:55 minutes and 
with the old version where I do not search for indirect inheritance is 18:45 minutes.
```

@martinvw, I have talked with @tanyatanyatanya and we have identified three options:

- revert the PR to the initial version;
- continue the work on the current one addressin the comments that are left;
- close the PR;

We would like to hear your opinion on that.

dstivanov:
@martinvw Seems that this PR progress depends on your opinion for the approach to be used. See the comment above. Can you please share your point of view?

martinvw:
I took so long hoping I did not have to have the final call.

I really think that limited validation a check like this can offer is not worth 1:36 and definitely not 3:44. 

I strongly think we should really keep this (build time vs added 'protection') always on our radar. 

An alternative could be that we apply it just as Sonar, so not run it in the build by default (or maybe only the error/warning checks) and when you run it explicitly you are aware that it will take longer.

Disclaimer I'm not the person requesting this check so it is hard for me to validate how much not having this check is going to hurt us.

kaikreuzer:
I am also ok to drop this check - we are rather looking at the API side and try to remove setters from our interfaces and rather offer people builders for modification of entities.

dstivanov:
@kaikreuzer I agree that the decision to alter the API to restrict incorrect access will be the best we can do here. 
As an outcome of the discussion I see a really good point in @martinvw 's opinion about the `build time vs added 'protection' `. If you find it reasonable, too - we can add some time measurements for the SAT execution and add such reports for the new checks in order to decide whether the build time taken will worth the gain.

kaikreuzer:
If a time logging can be easily added, that might make sense to get an idea of the impact of the different checks.

dstivanov:
This PR will be closed as the restrictions will be applied on API side instead of statically checked. @tanyatanyatanya , thank you a lot for the effort you put in here!

PR #224, opened 2017-11-30
Added check to look for proper contribution description of javadoc authors

### LINKED ISSUE #199: Enrich the AuthorTagCheck with "Initial contribution" Check for the First Author
Usually, the first mentioned author in the Java file has done the Initial contribution. Can we enrich the AuthorTagCheck in order to check whether the "Initial contribution" is present as the description of the contribution of the first author in the list? 

However, as the severity of the AuthorTagCheck is ERROR we can think of another check,  that will produce WARNING, if 

- There is no description for some of the authors
- There is not "Initial contribution" for the first author.

kssimova:
Checks if there is proper contribution description of the first javadoc author and random(just not empty) contribution description of every other javadoc author and generates a warning if they are missing.

Closes #199 

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
@@ -214,6 +214,12 @@
       <property name="severity" value="info"/>
       <property name="allowMissingPropertyJavadoc" value="true"/>
     </module>
+    
+    <module name="org.openhab.tools.analysis.checkstyle.AuthorContributionDescriptionCheck">
+       <property name="severity" value="warning"/>
+       <property name="requiredContributionDescriptions" value="initial contribution"/>
+       <property name="checkInnerUnits" value="false"/>
+    </module>
 
     <!-- Rules for Java Naming Convention -->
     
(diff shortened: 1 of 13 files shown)


src/test/java/org/openhab/tools/analysis/checkstyle/test/AuthorContributionDescriptionCheckTest.java
```diff
@@ -0,0 +1,283 @@
+package org.openhab.tools.analysis.checkstyle.test;
```
svilenvul: Could you add the java header ?

kssimova:
@kaikreuzer I updated this PR, it would be great if you take a look.

sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
```diff
@@ -214,6 +214,12 @@
       <property name="severity" value="info"/>
       <property name="allowMissingPropertyJavadoc" value="true"/>
     </module>
+    
+    <module name="org.openhab.tools.analysis.checkstyle.AuthorContributionDescriptionCheck">
+       <property name="severity" value="warning"/>
+       <property name="requiredContributionDescriptions" value="initial contribution,initial contribution and api"/>
```
kaikreuzer: Since you are checking with a "contains", it should be enough to list "initial contribution" here, since the other string is then automatically covered as well, right?

kssimova: Yes, you are absolutely right! Thanks!

PR #226, opened 2017-12-01
Use http URL to download .xsd files.

svilenvul:
Https connection causes 'javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure'
because https://www.eclipse.org is using TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 cypher which is not provided by standard JDK 1.8 and requires Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy Files 8 to be locally installed in JAVA_HOME

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
@@ -123,7 +123,7 @@
     <property name="severity" value="error" />
     <property name="thingSchema" value="http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd" />
     <property name="bindingSchema" value="http://www.eclipse.org/smarthome/schemas/binding-1.0.0.xsd" />
-    <property name="configSchema" value="https://www.eclipse.org/smarthome/schemas/config-description-1.0.0.xsd" />
+    <property name="configSchema" value="http://www.eclipse.org/smarthome/schemas/config-description-1.0.0.xsd" />
   </module>
   
   <module name="org.openhab.tools.analysis.checkstyle.EshInfXmlUsageCheck">


dstivanov:
It's strange how we haven't hit that before... Let's see @kaikreuzer opinion on  that, because I guess that the problem is related to the EPL to EPL v.2 change.

svilenvul:
Actually I mislead you a bit :-) , the Exception was thrown while I was trying to change the [following line](https://github.com/openhab/static-code-analysis/blob/3406e40c388f23530a307f31fc23accd192c4559/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/utils/CachingHttpClient.java#L75)
 from
`URLConnection connection = url.openConnection();`
to
`HttpsURLConnection connection = (HttpsURLConnection)url.openConnection();`

As this change was not possible because of the reason I pointed above, I switched back to the http URL.

svilenvul:
I forgot to mention that this error was produced after I tried to change the type of the connection to https [here](https://github.com/openhab/static-code-analysis/blob/3406e40c388f23530a307f31fc23accd192c4559/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/utils/CachingHttpClient.java#L75):

from
`URLConnection connection = url.openConnection();`
to
`HttpsURLConnection connection = (HttpsURLConnection) url.openConnection();`

Maybe this is why we haven't seen it before. But anyway I didn't find a way to establish https connection to  https://www.eclipse.org without the  Java Cryptography Extension (JCE).

dstivanov:
It's strange how we haven't hit that before... Let's see @kaikreuzer opinion on  that, because I guess that the problem is related to the EPL to EPL v.2 change.

kaikreuzer:
I have no idea why we haven't seen this before - maybe Eclipse updated their certificates on the server. But the change is fine for me, we do not require https here.

PR #236, opened 2017-12-22
Changed the default manifest version of PomXmlCheck

tanyatanyatanya:


svilenvul:
@kaikreuzer,  

keep in mind that this check has a severity error, so it is a build breaking.

PR #237, opened 2017-12-22
Added detector for mismatching placeholder-parameter in logger

### LINKED ISSUE #213: [Logging] add more FindBugs patterns from findbugs-slf4j
The tool is using https://github.com/KengoTODA/findbugs-slf4j for verifying logging messages, but not all patterns that this plugin offers are in use.

I have noticed two good candidates:
- [this pattern alerts if count of placeholder does not match to count of parameter](https://github.com/KengoTODA/findbugs-slf4j#slf4j_place_holder_mismatch)
- [this pattern reports that illegal class is passed to LoggerFactory.getLogger(Class)](https://github.com/KengoTODA/findbugs-slf4j#slf4j_illegal_passed_class)

In order to add them to the tool, the detectors that recognize these patterns should be added to the tool configuration.

I think the severity is currently error for FindBugs checks, so the code base might have to be cleaned up.

kssimova:
Closes #213 

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/findbugs/exclude.xml
+++ b/sat-plugin/src/main/resources/rulesets/findbugs/exclude.xml
@@ -24,4 +24,10 @@
     <Class name="~.*Util"/>
     <Bug pattern="SLF4J_LOGGER_SHOULD_BE_NON_STATIC"/>
   </Match>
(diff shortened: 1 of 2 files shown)


martinvw:
LGTM

PR #238, opened 2018-01-08
Add check to generate a warning if the exception is not added as a last parameter

### LINKED ISSUE #221: [Logging] add exception as last parameter
Another possible logging check:
- Where ever unchecked exceptions are caught and logged, the exception should be added as a last parameter to the logging
- For checked exceptions, this is normally not recommended, unless it can be considered an error situation and the stacktrace would hold additional important information for the analysis.

ghost:
Add new check to generate a warning if the exception is not added as a last parameter to the logging for caught and logged unchecked exceptions.
Fixed #221

### DIFF
--- a/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ExceptionLastParameterCheck.java
+++ b/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ExceptionLastParameterCheck.java
@@ -1,0 +1,229 @@
+/**
+ * Copyright (c) 2010-2018 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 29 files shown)


svilenvul:
@martinvw, please take a look when you have time for that.

dstivanov:
@tanyatanyatanya Can you please adapt the PR according to the changes in #287 ?

tanyatanyatanya:
@dstivanov Done.

Hilbrand:
There is a pmd check that kind of checks for this too: [UseCorrectExceptionLogging](https://pmd.github.io/latest/pmd_rules_java_errorprone.html#usecorrectexceptionlogging). Was that considered or doesn't cover the use case implemented in this check?

kaikreuzer:
Closing due to inactivity.

PR #241, opened 2018-01-10
Do not allow any e.printStackTrace() anymore.

martinvw:
See also https://github.com/openhab/openhab2-addons/pull/3104

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/pmd/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/pmd/rules.xml
@@ -53,10 +53,10 @@
     <priority>3</priority>
   </rule>
   <rule ref="rulesets/java/logging-java.xml/AvoidPrintStackTrace">
-    <priority>3</priority>
+    <priority>1</priority>
   </rule>
   <rule ref="rulesets/java/logging-java.xml/SystemPrintln">
-    <priority>3</priority>
+    <priority>2</priority>
   </rule>
   
   <!-- See http://pmd.sourceforge.net/pmd-5.0.0/rules/java/strictexception.html -->


sat-plugin/src/main/resources/rulesets/pmd/rules.xml
```diff
@@ -53,10 +53,10 @@
     <priority>3</priority>
   </rule>
   <rule ref="rulesets/java/logging-java.xml/AvoidPrintStackTrace">
-    <priority>3</priority>
+    <priority>1</priority>
```
dstivanov: Have you tested it on openHAB-addons and ESH builds? Are we sure that it will not break the build ?

martinvw: I created a PR for openhab, have to check for ESH

martinvw: Hmm, I have bigger problems when building openhab-core 

```
[ERROR] org.openhab.core.internal.icon.IconForwarder.java:[2]
Header line doesn't match pattern ^ \* Copyright \(c\) 2010-2018 by the respective copyright holders\.$
```

dstivanov: Yes, this is due to the recent pull request from @tanyatanyatanya, that was merged yesterday... So we have to fix the repositories before using this SAT version on them.

lpapazow: I think the reason for these problems is that the copyright headers in `openhab-core` start with 
> Copyright (c) 2015-2018

but the copyright headers in `openhab2-addons` and the expected header pattern in the `static-code-analysis` tool are
 > Copyright (c) 2010-2018

I will add a `ruleset.properties` file in openhab-core that overrides the expected copyright header pattern for the openhab-core repo and create a PR.

svilenvul: Already opened, waiting for the new release : https://github.com/openhab/openhab-core/pull/262

PR #256, opened 2018-02-16
fix 'empty line after header' false positive warning for some headers

lpapazow:
Fix lines that are resolved as headers because of their following line such as:

> Alt-H1
======

incorrectly triggering **Missing an empty line after the Markdown header**  warning


svilenvul:
LGTM :)

PR #263, opened 2018-02-26
Modified MethodNameCheck's regular expression to match custom OSGi (Equinox) console commands

kssimova:
Modified MethodNameCheck's regular expression to match methods which begin with underscore (custom OSGi console commands) and fixed AuthorContributionDescriptionCheckTest's warning message.

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
@@ -217,7 +217,7 @@
     
     <module name="org.openhab.tools.analysis.checkstyle.AuthorContributionDescriptionCheck">
        <property name="severity" value="warning"/>
-       <property name="requiredContributionDescriptions" value="initial contribution"/>
+       <property name="requiredContributionDescriptions" value="Initial contribution"/>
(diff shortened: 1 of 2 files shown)


sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
```diff
@@ -234,6 +234,8 @@
       <property name="severity" value="info" />
     </module>
     <module name="MethodName">
+        <property name="format"
+            value="^\_?[a-z][a-zA-Z0-9]*$"/>
```
martinvw: Should we just keep those info's as know defects to always 'promote' underscore feels a little bit to broad

kssimova: Should we make them warnings then? Or errors? The other option, which will not 'promote' underscore is to suppress them (only 2 files).

martinvw: No, info is fine, but just not allow the underscore and indeed suppressing for those two files sounds more sane then changing the regex.

Issue #255, opened 2018-02-06
Check reports files that are automatically generated


kssimova:
Currently, *NewLineAtEndOfFileCheck* reports **.xml** files, which are located in the **OSGI-INF** folder and are included in *.gitignore* file. The check is from *com.puppycrawl.tools.checkstyle.checks*, so changing it directly is not an option. As these files are automatically generated, what should we do to not report them anymore? Perhaps we should find a way to ignore all **OSGI-INF** folders for the check? 
@kaikreuzer @martinvw wdyt?

tanyatanyatanya:
@kssimova  OnlyTabIdentationInXmlFilesCheck also reports .xml files located in the OSGI-INF folders.

lpapazow:
I will create a custom checkstyle [filter](http://checkstyle.sourceforge.net/writingfilters.html) that checks if a file is ignored by git (a .gitignore file is present in the same directory) and skips the file if it is .gitignored.

The filter should be generic and suppress reports for all checks.

PR #268, opened 2018-03-07
Changed the severity from warning to error for OnlyTabIdentationInXmlFilesCheck

tanyatanyatanya:
@kaikreuzer @maggu2810 As we talked, after fixing all XML files found by the Static Code Analysis Tool for the OnlyTabIdentationInXmlFilesCheck I changed the severity of the check from warning to error.

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
@@ -76,7 +76,7 @@
   </module>
   
   <module name="org.openhab.tools.analysis.checkstyle.OnlyTabIndentationInXmlFilesCheck">
-    <property name="severity" value="warning" />
+    <property name="severity" value="error" />
     <property name="onlyShowFirstWarning" value="true" />
   </module>
 
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/suppressions.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/suppressions.xml
@@ -24,4 +24,5 @@
     <!--  Allow the usage of scheduleAtFixedRate in FadingWiFiLEDDriver class  -->
     <suppress files=".+org.openhab.binding.wifiled.handler.FadingWiFiLEDDriver.java" checks="AvoidScheduleAtFixedRateCheck"/>
     <suppress files=".+[\\/]pom\.xml" checks="OnlyTabIndentationInXmlFilesCheck"/>
+    <suppress files=".+[\\/]OSGI-INF[\\/]org.openhab.+\.xml" checks="OnlyTabIndentationInXmlFilesCheck|NewlineAtEndOfFileCheck"/>
 </suppressions>


martinvw:
Was it also done in the openHAB repo's?

tanyatanyatanya:
No, I will fix this right away.Thanks for the reminder.

svilenvul:
@tanyatanyatanya, do we have any progress on this one ?

martinvw:
I actually think @wborn fixed them :-)

svilenvul:
OK, thanks ! @tanyatanyatanya, if all repos are clean, please rebase, and we are ready to go :).

I see some commits in openhab-core as well - https://github.com/openhab/openhab-core/commit/8a3b3d10742562ed2b9b9e0878ea43f7f4ce5f11

wborn:
> I actually think @wborn fixed them :-)

That's right!

As a developer you will still find occurrences of the OnlyTabIdentationInXmlFilesCheck when you make a local build because Eclipse generates XML files from DS annotations. So either we should find a way to make the generated XML comply to the checks or exclude these generated XML files from SAT checks.

This is an example of such a generated XML file (`/openhab2-addons/addons/binding/org.openhab.binding.chromecast/OSGI-INF/org.openhab.binding.chromecast.internal.ChromecastDiscoveryParticipant.xml`):
```
<?xml version="1.0" encoding="UTF-8"?>
<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" immediate="true" name="org.openhab.binding.chromecast.internal.ChromecastDiscoveryParticipant">
   <service>
      <provide interface="org.eclipse.smarthome.io.transport.mdns.discovery.MDNSDiscoveryParticipant"/>
   </service>
   <implementation class="org.openhab.binding.chromecast.internal.ChromecastDiscoveryParticipant"/>
</scr:component>
```

These files are not committed to repositories because they match a .gitignore filter. So perhaps a generic workaround could be to exclude any file from SAT checks that is ignored due to .gitignore files?

maggu2810:
@wborn AFAIK The XML files that are created by the Maven build process (that also triggers SAT) are generated in the directory `${project.build.outputDirectory}/OSGI-INF` (normally `target/classes/OSGI-INF`). The build directory should be not inspected by SAT.
Do the Eclipse IDE generate that files into the "normal" OSGI-INF directory? That's weird.
But shouldn't it be enought to call `git clean -x -d -f`` in front of the Maven build (IMHO this is a good practice because you can be sure that no other files slip into the build that shouldn't be present).

tanyatanyatanya:
@wborn @maggu2810 There is already an open issue about these files that are automatically generated and they match a .gitignore filter #255 by @kssimova. 
And perhaps it is a good idea to exclude any file from SAT checks that is ignored due to .gitignore files as suggested @wborn

tanyatanyatanya:
@wborn The build will fail.

kssimova:
@maggu2810 @wborn I tried a few times with `git clean -x -d -f` right before the Maven build, but some XML files, which are generated from DS annotations, are still reported by *NewlineAtEndOfFileCheck* and *OnlyTabIndentationInXmlFilesCheck*.

wborn:
@kssimova Some generated files may have been committed as workaround for https://github.com/openhab/openhab-core/issues/318. When Eclipse is still running it may also regenerate the files. 

We could also add a suppression filter for these generated files. Depending on the project it could suppress these checks for:
* `.+[\\/]OSGI-INF[\\/]org.eclipse.smarthome.+\.xml`
* `.+[\\/]OSGI-INF[\\/]org.openhab.+\.xml`

It's also easy to forget the git clean command, which would result in a failing build when the severity is raised to error. :-(

svilenvul:
I personally like the idea of adding suppressions as @wborn proposed. 

The current situation is still messy because part of the DS files are generated (should be skipped) and other are still checked out in the repository (should be checked). Adding suppression filter only for the generated one looks good.

@tanyatanyatanya or @kssimova, could you update the PR with the proposed changes and rebase ?

svilenvul:
Thanks !

PR #274, opened 2018-03-20
Add parameter "includes" in pmd.properties file

tanyatanyatanya:
After fixing one of the issues found by Static Code Analysis Tool on smarthome repository for  BuildPropertiesCheck, by adding src/main/java/ and src/test/java/ directories to "source" property in build.properties file in /smarthome/bundles/test/org.eclipse.smarthome.test, I got an error from PMD for one of the test files - "org.eclipse.smarthome.test.java.JavaTestTest.java".

The test file is written by @htreu and the error is for "throwing null pointer exception", so I added parameter "includes" in pmd.properties, so only files in src/main/java/ will be checked.

@kaikreuzer @martinvw What do you think? Should we exclude the test files for PMD?

### DIFF
--- a/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/PmdChecker.java
+++ b/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/PmdChecker.java
@@ -78,6 +78,7 @@ public class PmdChecker extends AbstractChecker {
         // so they have to be set direct in the configuration
         Xpp3Dom configuration = configuration(
                 element("targetDirectory", userProps.getProperty("pmd.custom.targetDirectory")),
+                element("includes", userProps.getProperty("pmd.custom.includes")),
                 element("rulesets", element("ruleset", rulesetLocation)));
 
         pmdPlugins.add(dependency("org.openhab.tools.sat.custom-checks", "pmd", plugin.getVersion()));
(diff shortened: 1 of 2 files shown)


svilenvul:
@tanyatanyatanya, looking at the [checkstyle configuration](https://github.com/openhab/static-code-analysis/blob/master/sat-plugin/src/main/resources/configuration/checkstyle.properties), I see that we have added more directories to be checked.

Maybe it makes sense to be consistent and apply the same configuration for PMD and FindBugs

htreu:
Since the `JavaTestTest` is only a test and the thrown NPE does nothing then produce an "always failing assertion" we could simply refactor it to throw an IllegalStateException. No need to exclude it from the checks. I will create a PR.

htreu:
Unfortunately it has to be an NPE since the JavaTest expects it during `waitForAssert`. The alternative is throwing an Error but thats equally bad style.
What about excluding just this line by adding `// NOPMD` comment to it?

tanyatanyatanya:
@svilenvul I guess we could apply the same configuration for PMD and FindBugs.

@htreu For now we will not check the test files. Furthermore if we apply the same configuration for PMD and FindBugs from the link that @svilenvul gave in the first comment, I do not see we are running the checks for test files.The current problem was that for PMD we did not have "includes" or "excludes" parameters.

htreu:
@tanyatanyatanya alright, thanks for clarification.

dstivanov:
@svilenvul Do you think that it makes sense to include all the files for PMD - markdown, Manifest, XML, properties, about.htm? My argument not to do it is the small amount of PMD checks (only one), related to source code only.

svilenvul:
I agree, the PMD checks are currently targeted on Java files, so this should be enough.

Issue #276, opened 2018-03-21
annotated optional arguments could break SAT


maggu2810:
Using annotated optional arguments is valid but breaks the SAT execution.
A simple setup demonstration could be found on [optional-arguments].

[optional-arguments]: https://github.com/maggu2810/static-code-analysis-test/tree/master/optional-arguments

```text
[ERROR] Failed to execute goal org.openhab.tools.sat:sat-plugin:0.4.1:checkstyle (default) on project optional-arguments: Unable to execute mojo: An error has occurred in Checkstyle report generation. Failed during checkstyle configuration: Exception was thrown while processing /home/rathgeb/workspace/projects-oss/static-code-analysis-test/optional-arguments/src/main/java/tmp/App.java: MismatchedTokenException occurred during the analysis of file /home/rathgeb/workspace/projects-oss/static-code-analysis-test/optional-arguments/src/main/java/tmp/App.java. expecting EOF, found '}' -> [Help 1]
```

maggu2810:
See also checkstyle/checkstyle#5629

maggu2810:
@svilenvul How can I add and configure the `BeforeExecutionExclusionFileFilter`?
Does SAT also support to modify the configuration or the suppressions only?

PR #278, opened 2018-03-22
add support to use checkstyle's BeforeExecutionExclusionFileFilter

maggu2810:
Related to: https://github.com/openhab/static-code-analysis/issues/276


maggu2810:
@svilenvul How does the release roadmap looks like?
I need it for eclipse/smarthome#5293 as without that I cannot add the correct nullness annotations without breaking the build.

svilenvul:
We would like to prepare a new release, but first we have to carefully test the latest state of the tool on the ESH/openHAB repositories.

PR #280, opened 2018-03-26
PMD execution bug fix

VelinYordanov:
After #274 PMD doesn't execute at all. This should fix the problem.


svilenvul:
@wborn, thanks for the feedback.

PR #281, opened 2018-03-26
Added check to look for proper null annotations

### LINKED ISSUE #218: Null annotations check
From the ESH coding guidelines:

- [x] - every bundle should have an optional Import-Package dependency to `org.eclipse.jdt.annotation`;
- [ ] - Classes should be annotated by @NonNullByDefault ;
- [ ] - return types, parameter types, generic types etc. are annotated with @Nullable only. There is no need for a @NonNull annotation because it is set as default;

kssimova:
Closes #218 


kssimova:
@svilenvul Done :)

svilenvul:
Thanks !

PR #290, opened 2018-04-11
Add Eclipse project files to .gitignore

wborn:
I've added the files Eclipse generates for projects to .gitignore so they are ignored when committing files.


svilenvul:
Thanks !

PR #292, opened 2018-04-13
create checkstyle filter and checks for TrailingWhitespaces and Windo…

### LINKED ISSUE #272: Trailing whitespaces
WDYT about a test that complains about trailing whitespace?
An inclusion and exclusion list (regular expression for the path pattern) could make sense.

### LINKED ISSUE #273: Standardize line endings using checks or .gitattributes
Currently the code bases contains a mix of both Unix and Windows line endings.

There are several ways to standardize the line endings being used:
* It can be done by adding .gitattributes files in your repo so they are converted automatically. See also the [documentation](https://git-scm.com/docs/gitattributes) and some commonly used [templates](https://github.com/alexkaratarakis/gitattributes).
* We could also add checks for it in SAT.

On the command line you can find files with Windows line endings with find using:
```
find . -not -type d -not \( -path "*target/*" -prune \) -not \( -path "*.git/*" -prune \) -not \( -path "*.settings/*" -prune \) -not \( -path "*/node_modules/*" -prune \) -not \( -path "*.cmd" -prune \) -exec file "{}" ";" | grep CRLF
```

lpapazow:
The checkstyle filter sets max number of violations reported.

The PR also includes WindowsLineEndingsCheck which fixes #273
and TrailingWhitespacesCheck  which fixes #272 

There are currently no tests because the current testing logic allows only classes that inherit `AbstractCheck` can be tested and `RegexpSinglelineCheck` inherits `AbstractFileSetCheck`. (see #294) 

Testing these rules should be possible when #296 is merged

### DIFF
--- a/custom-checks/checkstyle/src/test/java/org/openhab/tools/analysis/checkstyle/test/AuthorContributionDescriptionCheckTest.java
+++ b/custom-checks/checkstyle/src/test/java/org/openhab/tools/analysis/checkstyle/test/AuthorContributionDescriptionCheckTest.java
@@ -28,7 +28,7 @@ import com.puppycrawl.tools.checkstyle.utils.CommonUtils;
  */
 public class AuthorContributionDescriptionCheckTest extends AbstractStaticCheckTest {
 
-    private static final String EXPECTED_WARNING_MESSAGE_FIRST_AUTHOR_DESCRIPTION = "First javadoc author should have \"Initial contribution\", \"Initial contribution and API\" contribution description.";
+    private static final String EXPECTED_WARNING_MESSAGE_FIRST_AUTHOR_DESCRIPTION = "First javadoc author should have \"Initial contribution\" contribution description.";
     private static final String EXPECTED_WARNING_MESSAGE_OTHER_AUTHOR_DESCRIPTION = "Javadoc author should not have empty contribution description.";
     private static final String TEST_DIRECTORY_NAME = "authorContributionDescriptionCheckTest";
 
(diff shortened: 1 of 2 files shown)


svilenvul:
Waiting for tests after #296 is merged.

wborn:
Did you also have a look at using a .gitattributes file instead of adding the WindowsLineEndingsCheck  @lpapazow? 

I agree with the comment by @kaikreuzer (https://github.com/openhab/static-code-analysis/issues/273#issuecomment-381090643) that using a .gitattributes file would cause fewer annoyances because git should convert them seamlessly. 

The GitHub [setup documentation](https://help.github.com/articles/dealing-with-line-endings/#platform-all) also recommends this approach.

I haven't used it myself though. So I don't know what kind of side effects it can cause. I can imagine that copying files between a Windows checkout and a Linux openHAB instance could cause issues (and vice versa). Though that would currently already be an issue with the mixed line endings. Another benefit would be that SAT does not have to do the check so it takes less time to run it.

lpapazow:
I thought that such a check could be used to validate that adding the .gitattributes file in the smarthome repo changes the eol symbols properly.

I added .gitattributes file in my local repo with attribute `eol=lf`.
After that when using git add I get the following warning:

`warning: CRLF will be replaced by LF in <name of file>
The file will have its original line endings in your working directory.`

Therefore, I agree with this [#273 comment](https://github.com/openhab/static-code-analysis/issues/273#issuecomment-381090643) that it would indeed be confusing and counter-productive.

If you also agree that this check won't be beneficial, I will just remove it from the PR. @wborn wdyt?

PR #295, opened 2018-04-16
Initial contribution of the ManifestLineLengthCheck

### LINKED ISSUE #285: manifest max. line length
We should check the line length of a manifest to fit the requirements.

**Notes on Manifest and Signature Files** 

Line length:
No line may be longer than 72 bytes (not characters), in its UTF8-encoded form. If a value would make the initial line longer than this, it should be continued on extra lines (each starting with a single SPACE).

https://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html

VelinYordanov:
Closes #285


VelinYordanov:
@maggu2810 any remarks?

Issue #294, opened 2018-04-13
Only checks that inherit from AbstractCheck can be tested by the AbstractStaticCheckTest logic


lpapazow:
Check classes that inherit from `AbstractFileSetCheck`, such as `RegexpSinglelineCheck` can't be tested.

Trying to test such class results in the following error: 

> com.puppycrawl.tools.checkstyle.api.CheckstyleException: cannot initialize module com.puppycrawl.tools.checkstyle.TreeWalker - TreeWalker is not allowed as a parent of org.openhab.tools.analysis.checkstyle.TrailingWhitespacesCheck Please review 'Parent Module' section for this Check in web documentation if Check is standard.
	at com.puppycrawl.tools.checkstyle.Checker.setupChild(Checker.java:461)
	at com.puppycrawl.tools.checkstyle.api.AutomaticBean.configure(AutomaticBean.java:171)
	at com.puppycrawl.tools.checkstyle.BaseCheckTestSupport.createChecker(BaseCheckTestSupport.java:81)
	at com.puppycrawl.tools.checkstyle.BaseCheckTestSupport.verify(BaseCheckTestSupport.java:239)
	at org.openhab.tools.analysis.checkstyle.test.TrailingWhitespacesCheckTest.verifyJavaFile(TrailingWhitespacesCheckTest.java:33)
	at org.openhab.tools.analysis.checkstyle.test.TrailingWhitespacesCheckTest.verifyScheduleAtFixedRateUsed(TrailingWhitespacesCheckTest.java:24)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
[...]

Issue #225, opened 2017-12-01
Update Checkstyle API


svilenvul:
Currently we are using 8.1, while 8.5. is released.

Looking at the [release notes](http://checkstyle.sourceforge.net/releasenotes.html#Release_8.5), I see some braking API changes that will affect us... :
- Disallowing user to use incomplete fully qualified Check names in config file;
- BaseCheckTestSupport is deprecated;

Most of the changes are bug fixes, I don`t see many new checks (new Check: single line annotation location), so maybe we can wait a bit with the update.

svilenvul:
@kssimova mentioned that some false positives in the checks for javadoc are fixed in the newer checkstyle realaeses. Could you provide more details on that ?

kssimova:
**UnusedImportsCheck** reports imports as unused, even though they are used in a Javadoc. For example: 
```java
import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;

/**
 * The normalizer for the {@link ConfigDescriptionParameter.Type#TEXT}. It basically ensures that the given value will
 * turned into its {@link String} representation.
 *
 * @author Simon Kaufmann - initial contribution and API.
 * @author Thomas Höfer - made class final and minor javadoc changes
 */
final class TextNormalizer extends AbstractNormalizer {

    @Override
    public Object doNormalize(Object value) {
        if (value == null) {
            return value;
        }
        if (value instanceof String) {
            return value;
        }
        return value.toString();
    }

}

```
I did some research and found that this problem is fixed in **Checkstyle** 8.3 : https://github.com/checkstyle/checkstyle/issues/3453, while we're currently using 8.1.

svilenvul:
Definitely not the single issue being fixed, but the first one that observed in the reports for ESH/openHAB.

PR #296, opened 2018-04-17
update checkstyle version to 8.9

lpapazow:
Update the checkstyle version from 8.1 to 8.9.

This should provide a solution for #294 and #225.

Looking at the [release notes](http://checkstyle.sourceforge.net/releasenotes.html#Release_8.9) these API-breaking changes that could potentially affect us are resolved :

* *Disallowing user to use incomplete fully qualified Check names in config file.*
We are using just the check name, not the fully qualified name, so this check should not affect us.
e.g: We use `NewlineAtEndOfFile`  in our configuration file and currently both `NewlineAtEndOfFile` and `com.puppycrawl.tools.checkstyle.checks.NewlineAtEndOfFileCheck` are allowed. 
What is no longer allowed is `com.puppycrawl.tools.checkstyle.checks.NewlineAtEndOfFile`

* *BaseCheckTestSupport is deprecated.*
All test classes are refactored to compile and work as intended.

* *Split TreeWalker to TreeWalker and JavaParser.*
JavaParser is a utility class with static methods that contains methods that were previously in the TreeWalker class. We do not use these methods directly.

### DIFF
--- a/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheckTest.java
+++ b/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/api/AbstractStaticCheckTest.java
@@ -9,10 +9,9 @@
 package org.openhab.tools.analysis.checkstyle.api;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.ArrayList;
 
-import com.puppycrawl.tools.checkstyle.BaseCheckTestSupport;
+import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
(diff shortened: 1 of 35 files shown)


svilenvul:
LGTM.

Wow, it is great that we didn't have to change the existing checks implementations (only the tests).

Waiting for @lpapazow to confirm that there is no regression in ESH/openHAB builds.

lpapazow:
I can confirm that there is no regression for ESH, OpenHab-Core and OpenHab2-addons. The only difference in checkstyle 8.9 is that some false positives for UnusedImportsCheck are gone.

svilenvul:
Great, thanks !

PR #300, opened 2018-05-02
create a check to warn if there are trailing Whitespaces

### LINKED ISSUE #272: Trailing whitespaces

lpapazow:
This PR fixes #272.

Currently, it's not possible to create tests because of #294.

Once #296 is merged tests could be created.

### DIFF
--- a/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/TrailingWhitespacesCheck.java
+++ b/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/TrailingWhitespacesCheck.java
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) 2010-2018 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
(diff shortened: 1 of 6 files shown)


svilenvul:
The update is merged, so you can continue with the tests.

lpapazow:
Tests are now included.

martinvw:
Is this check really worth it, I don't know what the added time on ESH / OH is for this check, but I personally care much more about build time then whitespace on the end of my lines, wdyt?

martinvw:
@maggu2810, what about my remark above, checking every single line of every single file with a regex does not sound like a free check for me?

maggu2810:
Hm, personally I differ between
* I would like to execute the checks and grant that the sources fits our requirements
* I would like to get new build artifacts (and don't execute the tests)

I would assume that most of the checks that operate on the source code files and not on the bytecode read and inspect every line of the file.
So, if you really want to improve the situation, checks should be combined to not read every file again and again...
But I don't think that "check for whitespaces" consumes more time as every other check that inspect the source code.

lpapazow:
With the addition of #316 it will be possible to only perform analysis on the changed bundles. Maybe with this option the check would be worth the time?
@martinvw wdyt?

kaikreuzer:
Closed as suggested in https://github.com/openhab/static-code-analysis/issues/272#issuecomment-619452218

Issue #304, opened 2018-05-16
Hard-coded default value for manifestVersionRegularExpression / currentVersionRegex


pfink:
We have a hard-coded value for manifestVersionRegularExpression / currentVersionRegex [here](https://github.com/openhab/static-code-analysis/blob/master/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml#L150).

This is a very bad smell as it will immediately break the code style check after a version bump. @VelinYordanov @svilenvul: Any suggestion for resolution? Or am I wrong?

svilenvul:
Have you tried to add a new property to a checkstyle.properties file like this one - https://github.com/openhab/openhab-core/blob/master/tools/checkstyle.properties :

```
checkstyle.headerCheck.content=^/\\*\\*$\\n^ \\* Copyright \\(c\\) {0}-{1} by the respective copyright holders\\.$\\n^ \\*$\\n^ \\* All rights reserved\\. This program and the accompanying materials$\\n^ \\* are made available under the terms of the Eclipse Public License v1\\.0$\\n^ \\* which accompanies this distribution, and is available at$\\n^ \\* http://www.eclipse.org/legal/epl\\-v10\\.html$
checkstyle.headerCheck.values=2015,2018 
checkstyle.pomXmlCheck.currentVersionRegex= add_your_regex_here
```

dstivanov:
@pfink The regex is taken from `${checkstyle.pomXmlCheck.currentVersionRegex}` and as it is always needed with added a default value 2.3.0. So the suggestion of @svilenvul should work well. 
Please do not hesitate to contact us again if you have other problems.

pfink:
Yes, I know this solution and [I already use it for my binding](https://github.com/pfink/openhab2-flicbutton/blob/master/tools/checkstyle.properties#L3). But it only moves the problem to another location, it is still a hard-coded value that needs extra maintenance on every version bump. Currently, it seems to be undocumented that this value has to be changed on every release and it could be easily forgetten. Also, the value is not overriden on our official repos (e.g. [here](https://github.com/openhab/openhab2-addons/blob/master/tools/checkstyle.properties)). So for them, currently the default value is used which means:
- we always have to create a static-code-analysis release for each openHAB GA release (every time when there is a snapshot version bump)
- we should integrate this version bump into our automated openHAB release process (otherwise it can be easily forgotten)
OR
- we find a better solution where this version is dynamically determined from the POM version

[...]

PR #306, opened 2018-05-22
Removed regex version properties from the PomXmlCheck

VelinYordanov:
See #304

With this change the check just verifies that the versions in the pom.xml and MANIFEST.MF are the same. **It can no longer check for a particular version.**


pfink:
Thanks, I think that makes sense! To have the same level of quality check as before, we could may walk up the POM hierarchy and check that the parent POMs have the same versions as the children. But we would need a property where this additional check can be easily turned off (because the assumption that parent POMs have the same version as child POMs valid for all of our openHAB projects, but not always for custom add-ons).

What do you think?

dstivanov:
@VelinYordanov Really, what do you think ?

kaikreuzer:
Manually merged with https://github.com/openhab/static-code-analysis/commit/49e6341d7380c7cec1e8e5ea8fe2f49053e1c85e

Issue #284, opened 2018-03-31
add "skip" configuration option


maggu2810:
The configuration of the mojo should support a "skip" parameter.
A skip parameter is supported by a lot of mojos.

This will e.g. allow to add the plugin configuration to the base pom in its default profile and just set the property in dependence of e.g. a special file etc.

Perhaps there should be multiple skip options to allow an easy to use and fine granulated control.

WRT to the current usage in ESH there could be one global skip option for the plugin that is controlled / set in respect of the "skipChecks" property. "skipChecks" could e.g. set "sat.skip" that will be used in the respective plugin configuration. If that skip is set no checks are executed.

IMHO not necessary at the moment as the plugin itself could execute the tests dependent of the presence of files itself:
Then there could be properties (if "skip" is not set), that defines the checks that are executed.
The presence of the `MANIFEST.MF` could e.g. control the value of "executeManifestMfChecks".
The presence of the `src/.../feature.xml` could e.g. control the value of "executeKarafFeatureChecks".

maggu2810:
Example usage: https://github.com/eclipse/smarthome/compare/master...maggu2810:sat-skip?expand=1

splatch:
I would apprecieate this option. Current build of Eclipse Smart Home and OpenHab takes quite long time (due to tycho) and skipping static code analysis which parses most, if not all, files is place where I look for biggest performance boost.
Currently getting developer feedback from custom ESH build takes simply too much time.

splatch:
Yeah, I spotted it and use it. It saves a lot of time during build.

svilenvul:
@VelinYordanov is working on this one.

dstivanov:
As the PR is updated - the task is again ready for review.

PR #307, opened 2018-05-23
Added new properties for skipping checks based on configuration.

VelinYordanov:
See #284

### DIFF
--- a/docs/maven-plugin.md
+++ b/docs/maven-plugin.md
@@ -88,6 +88,8 @@ Parameters:
 | **pmdFilter** | String | Relative path of a suppression.properties file that lists classes and rules to be excluded from failures. If not set no classes and no rules will be excluded |
 | **maven.pmd.version** | String | The version of the maven-pmd-plugin that will be used (Default value is **3.7**)|
 | **pmdPlugins** | List<Dependency> | A list with artifacts that contain additional checks for PMD |
+|**skip.pmd**|boolean|If set to true pmd checks will not be executed|
+|**skippedCategories**|Collection<String>|A collection of all categories that will be skipped during the execution of PMD. If not set all categories of checks will be executed. The categories are part of the official [PMD API](https://github.com/pmd/pmd/wiki/Rule-Categories).|
 
 ### sat-plugin:checkstyle
 
@@ -103,6 +105,8 @@ Parameters:
 | **maven.checkstyle.version** | String | The version of the maven-checkstyle-plugin that will be used (default value is **2.17**)|
 | **checkstylePlugins** | List<Dependency> | A list with artifacts that contain additional checks for Checkstyle |
 | **checkstyleProperties** | String | Relative path of the properties file to use in the ruleset to configure specific checks |
+|**skip.checkstyle**|boolean|If set to true no checkstyle checks will be executed.|
+|**skippedFileTypes**|Collection<String>|A collection of all the file types that will be skipped during the execution of checkstyle.If not set all checks are executed. If a check has multiple types and contains one of the skipped types it will be skipped.
 
 ### sat-plugin:spotbugs
 
@@ -120,6 +124,7 @@ Parameters:
 | **spotbugs.version** | String | The version of SpotBugs that will be used (default value is **3.1.1**)|
 | **spotbugsPlugins** | List<Dependency> | A list with artifacts that contain additional detectors/patterns for SpotBugs |
 | **findbugs.slf4j.version** | String | The version of the findbugs-slf4j plugin that will be used (default value is **1.2.4**)|
+|**skipSpotbugs**|boolean|If set to true all spotbugs checks will be skipped.|
 
 ### sat-plugin:report
 
@@ -170,7 +175,29 @@ The `visitors.xml` contains a list with SpotBugs visitors (bug detectors) and ha
   ...
 <visitors/>
 ```
+In order to skip certain types of checkstyle checks you need to add the ```skippedFileTypes``` property to the configuration or you can use ```-Dcheckstyle.skippedFileTypes``` The available types are : **java, xml, classpath, properties, markdown, manifest, html**:
+```
+<plugin>
+    <groupId>org.openhab.tools.sat</groupId>
+    <artifactId>sat-plugin</artifactId>
+    <version>${sat.version}</version>
+    <configuration>
+		<skippedFileTypes>manifest,properties</skippedFileTypes>
+    </configuration>
+</plugin>
+```
 
+In order to skip certain types of PMD categories you need to add the ```skippedCategories``` property to the configuration or you can add the categories with ```-Dpmd.skippedCategories```
+```
+<plugin>
+    <groupId>org.openhab.tools.sat</groupId>
+    <artifactId>sat-plugin</artifactId>
+    <version>${sat.version}</version>
+    <configuration>
+		<skippedCategories>design</skippedCategories>
+    </configuration>
+</plugin>
+```
 ### Individual plugin customization
 
 Each of the Maven plugins that are used (for SpotBugs, Checkstyle and PMD) are configured by setting user properties that are located in the `src/main/resources/configuration` directory.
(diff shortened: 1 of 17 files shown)


dstivanov:
@VelinYordanov Can you please rebase the PR in order to make it open for review ?

sat-plugin/src/main/java/org/openhab/tools/analysis/tools/CheckstyleChecker.java
```diff
@@ -127,4 +174,40 @@ public void execute() throws MojoExecutionException {
 
     }
 
+    private void generateRulesXmlFile(String location) throws MojoExecutionException {
+        try {
+            String rulesLocation = getLocation(checkstyleRuleset, DEFAULT_RULE_SET_XML);
```
martinvw: Are we using this method even if we do not skip anything, if I remember correctly XML manipulation like this is rather expensive, so please skip it if its not needed.

sat-plugin/src/main/java/org/openhab/tools/analysis/tools/PmdChecker.java
```diff
@@ -105,7 +152,34 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                 configuration, pmdPlugins);
 
         log.debug("PMD execution has been finished.");
-
     }
 
+    private void generateRulesXmlFile(String location) throws MojoExecutionException {
+        try {
```
martinvw: See above?

sat-plugin/src/main/java/org/openhab/tools/analysis/tools/PmdChecker.java
```diff
@@ -105,7 +152,34 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+            String rulesLocation = getLocation(pmdRuleset, DEFAULT_RULESET_XML);
+            URL url = new URL(rulesLocation);
+            SAXReader reader = new SAXReader();
+            Document document = reader.read(url);
+            List<Node> ruleNodes = document.getRootElement().elements("rule");
+            ruleNodes.forEach(node -> {
+                // There is another Element type coming from
+                // org.twdata.maven.mojoexecutor.MojoExecutor.*
```
martinvw: Then don't use a * import :-)

sat-plugin/src/main/java/org/openhab/tools/analysis/tools/PmdChecker.java
```diff
@@ -105,7 +152,34 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                        getLog().debug(String.format(CATEGORY_SKIP_MESSAGE, ruleType, category));
+                        // We detach the rule if it belongs to a skipped category
+                        node.detach();
+                    }
+                }
+            });
+
+            FileOutputStream stream = new FileOutputStream(location);
+            new XMLWriter(stream, OutputFormat.createPrettyPrint()).write(document);
+            stream.close();
```
martinvw: Are all streams properly closed in case of exceptions? Try-with-resources?

Should we close the reader or the document?

Of course also apply to the other one.

sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
```diff
@@ -40,39 +41,49 @@
...
   <module name="org.openhab.tools.analysis.checkstyle.OutsideOfLibExternalLibrariesCheck">
     <property name="severity" value="error" />
     <property name="ignoredDirectories" value="${checkstyle.outsideOfLibExternalLibrariesCheck.ignoredDirectories}" default="target,bin" />
+    <type value="properties"/>
   </module>
   
   <module name="org.openhab.tools.analysis.checkstyle.readme.MarkdownCheck">
     <property name="severity" value="warning" />
+    <type value="properties"/>
+    <type value="markdown"/>
```
martinvw: What is the consequence if one of the two types is requested to be skipped, should we document this behaviour?

docs/maven-plugin.md
```diff
@@ -103,6 +105,8 @@ Parameters:
 | **maven.checkstyle.version** | String | The version of the maven-checkstyle-plugin that will be used (default value is **2.17**)|
 | **checkstylePlugins** | List<Dependency> | A list with artifacts that contain additional checks for Checkstyle |
 | **checkstyleProperties** | String | Relative path of the properties file to use in the ruleset to configure specific checks |
+|**skipCheckstyle**|boolean|If set to true no checkstyle checks will be executed.|
+|**skippedCheckTypes**|Collection<String>|A collection of all the check types that will be skipped during the execution of checkstyle.If not set all checks are executed.
```
martinvw: To me this is not clear what these types are, while reviewing I realised it might be filetypes, is that so, should we spell that out here instead of check types?

sat-plugin/src/test/java/org/openhab/tools/analysis/tools/test/CheckstyleCheckerTest.java
```diff
@@ -0,0 +1,214 @@
...
+        Properties properties = mock(Properties.class);
+        CheckstyleCheckerMock checker = setUpExecutionCheck(properties);
+        PluginDescriptor plugin = mock(PluginDescriptor.class);
+        when(plugin.getVersion()).thenReturn("version");
+        checker.setPluginDescriptor(plugin);
+
+        try {
+            checker.execute();
+        } catch (Exception e) {
+            // We only want to see if the executeCheck method is called with the proper
```
martinvw: Is the exception expected? Because afaik a test will fail if it throws an exception, so if this test is green, you should be able to remove the try-catch-finally

martinvw: Apply everywhere

src/etc/rulesDTD.dtd
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
martinvw: What is this, where is it needed for? Tests? Then make it a test resource otherwise a normal resource?

docs/maven-plugin.md
```diff
@@ -88,6 +88,8 @@ Parameters:
 | **pmdFilter** | String | Relative path of a suppression.properties file that lists classes and rules to be excluded from failures. If not set no classes and no rules will be excluded |
 | **maven.pmd.version** | String | The version of the maven-pmd-plugin that will be used (Default value is **3.7**)|
 | **pmdPlugins** | List<Dependency> | A list with artifacts that contain additional checks for PMD |
+|**skipPmd**|boolean|If set to true pmd checks will not be executed|
+|**skippedCategories**|Collection<String>|A collection of all categories that will be skipped during the exeuction of PMD. If not set all categories of checks will be executed.|
```
martinvw: Are these categories something existing?

VelinYordanov: Yes, they are part of the official API https://github.com/pmd/pmd/wiki/Rule-Categories

docs/maven-plugin.md
VelinYordanov: Yes, they are the file types that the check processes. I added this to the docs.

sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
VelinYordanov: If one of the two types is requested to be skipped the check will be skipped. I added this to the docs.

sat-plugin/src/main/java/org/openhab/tools/analysis/tools/CheckstyleChecker.java
VelinYordanov: We can do this for pmd, but we can't for checkstyle, because the added types to the rules are not part of the official API. We have to at least remove them before executing.

martinvw: What is the (performance) impact of this?

sat-plugin/src/test/java/org/openhab/tools/analysis/tools/test/CheckstyleCheckerTest.java
VelinYordanov: If we want the exception to not be thrown a lot of mocking will be needed. I don't think it is necessary for testing the generated rules.xml file.

kaikreuzer:
Closing due to inactivity.

PR #311, opened 2018-05-28
Added configuration property for the OutsideOfLibExternalLibrariesCheck

### LINKED ISSUE #308: Refine the Exclusion of OutsideOfLibExternalLibrariesCheck
Due to the problems found in https://github.com/openhab/openhab2-addons/pull/3582 we need to rethink the OutsideOfLibExternalLibrariesCheck and the way we ignore files there. 

As I discussed with @kaikreuzer we will need a way to ignore specific files (not directories only) - for example a jar file, that is placed in the lib directory ("*_sources.jar" files are the most common example).

When the proposed configuration is released - please adapt the openhab2-addons to avoind using the workaround, proposed in https://github.com/openhab/openhab2-addons/pull/3582#issuecomment-392022273

VelinYordanov:
Closes #308


dstivanov:
It also looks good to me so I will merge the PR.

PR #312, opened 2018-05-28
Remove default suppressions.

VelinYordanov:
In my opinion if suppressions are needed they should be included in the respective repository. Having a default suppressions file in this repository seems confusing.
See openhab/openhab2-addons#3582


claell:
Just make sure this won't break anything. At least openhab2-addons will fail to build unless either the PR you referenced above gets merged or a new PR just for that is created.

There might be other repos that used this files before affected as well. Did you check that?

dstivanov:
@VelinYordanov Have you checked that for the other builds where SAT is used ?

Issue #232, opened 2017-12-15
Improve build times using the tool


svilenvul:
Currently the build times for openHAB and Eclipse SmartHome are very high (both for local build and CI PR builds) is and as the project grow this becomes even worse.

As we have already discussed in https://github.com/eclipse/smarthome/pull/3995#issuecomment-322392976 FindBugs(SpotBugs) analysis are especially slow.

What we have discussed with @dstivanov is, if we can make more intelligent use of the tool in the build system. We have gathered the following ideas:
- for nightly builds : use the current approach to execute all checks on all bundles;
- for PR CI builds :  search for an option to execute all checks only on bundles that are changed;
- for local builds : maybe do not execute the checks at all, or if possible execute only on changed bundles;

Most of the checks (except few FindBugs and a few exceptions) are isolated to a single bundle, so this should not be a compromise with the code quality aspect.

We wanted to start this discussion and together gather ideas, so please share your opinions on this topic.

svilenvul:
One way to accomplish the requirements above could be:

- write script that uses `git diff master` to see which files are being changed and extract the Maven artifact IDs of the changed bundles;
- add a new property to the static code analysis tool which accepts a list of Maven artifact IDs. In case the bundle is not in the list, the execution can be skipped.

@dstivanov, please ping me to discuss the details

svilenvul:
Checkstyle also has a [cache support](http://checkstyle.sourceforge.net/config_reporting.html#Caching_Support). Maybe additional configuration is needed in the ruleset as well - http://checkstyle.sourceforge.net/config.html

svilenvul:
Ideas can be gathered from SonarQube as well, maybe a good starting point is their issue tracker - [Support incremental code quality analysis](https://jira.sonarsource.com/browse/SONAR-2815?jql=text%20~%20%22incremental%22).

svilenvul:
For SpotBugs, incremental build support is in the 4.0.0 Roadmap - https://github.com/spotbugs/spotbugs/wiki/Roadmap. Unfortunately I can not see any progress on this feature, as there is no opened issue in GitHub.

The closest that I can find is https://github.com/spotbugs/spotbugs/issues/249

martinvw:
> I see that PMD supports Incremental Analysis since version 6.0.0 -https://pmd.github.io/pmd-6.2.0/pmd_userdocs_getting_started.html#incremental-analysis.
> 
[...]

PR #316, opened 2018-06-26
Create two more MOJOs that allow only changed modules to be analysed

lpapazow:
One of the steps suggested in #232 to improve build time.

Create two more static-analysis-tool goals and modify the existing ones to allow only bundles, in which there were changes to be checked thus saving a lot of time.

The first goal "create-scm-diff" is implemented in the ScmDiffMojo class. In this goal, "git diff master --names-only ./" is called to create a file, containing the names of all changed files in the module. This file is created for each module.

The second goal "find-changed-bundles" is implemented in the FindChangedBundlesMojo class. In this goal, for every changed file, the artifactId of the first parent pom is taken, and a maven property artifactIds is set that contains the comma-separated values of all the artifactIds.

These two goals should be executed in the package phase.

Later, when checkstyle, pmd and findbugs goals start executing in the verify phase, the code checks whether the current artifactId is contained in the artifactIds maven property and if it is contained, the goal is executed. If it is not contained, the bundle is not checked by checkstyle, pmd and findbugs.

In order for this to work, a separate profile in the smarthome pom needs to be added:

```<profile>
  <id>only-check-changed</id>
  <activation>
    <property>
      <name>onlyCheckChanged</name>
      <value>true</value>
    </property>
  </activation>
  <build>
    <pluginManagement>
      <plugins>
        <plugin>
          <groupId>org.openhab.tools.sat</groupId>
          <artifactId>sat-plugin</artifactId>
          <configuration>
            <diffOutputFile>target/smarthome.diff</diffOutputFile>
          </configuration>
          <executions>
            <execution>
              <id>create-diff-files</id>
              <goals>
                <goal>create-git-diff</goal>
                <goal>find-changed-bundles</goal>
              </goals>
              <phase>package</phase>
            </execution>
          </executions>
        </plugin>
      </plugins>
    </pluginManagement>
  </build>
</profile>
```
Additionally, In the project configuration, a scm configuration element must be added with the scm directory.

```
  <scm>
    <connection>scm:git:file:/.git</connection>
  </scm>
```

Also 
```<onlyCheckChanged>false</onlyCheckChanged>``` must be added as a property with default value false.

After adding these things to the smarthome pom configuration, the build should be executed with:
```mvn clean install -DonlyCheckChanged```. 

Signed-off-by:_ Lyubomir V. Papazov <lpapazow@gmail.com>

### DIFF
--- a/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/AbstractChecker.java
+++ b/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/AbstractChecker.java
@@ -13,6 +13,7 @@ import java.io.InputStream;
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Properties;
@@ -29,6 +30,7 @@ import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
+import org.openhab.tools.analysis.changes.FindChangedBundlesMojo;
 import org.twdata.maven.mojoexecutor.MojoExecutor;
 
 /**
@@ -44,6 +46,9 @@ public abstract class AbstractChecker extends AbstractMojo {
 
     @Parameter(defaultValue = "${session}", readonly = true)
     protected MavenSession mavenSession;
+
+    @Parameter(property = "onlyCheckChanged")
+    protected Boolean onlyCheckChanged;
 
     @Component
     private BuildPluginManager pluginManager;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: sat-plugin/src/main/java/org/openhab/tools/analysis/tools/AbstractChecker.java [hunk 1]
```diff
@@ -130,6 +135,30 @@ protected String getLocation(String externalRelativePath, String internalRelativ
...
+    protected boolean shouldAnalyseModule() {
+        if (!onlyCheckChanged) {
+            return true;
+        }
+
+        String artifactIds = mavenProject.getProperties().getProperty(FindChangedBundlesMojo.AFFECTED_ARTIFACT_IDS_PROPERTY);
+        if (artifactIds == null) {
+            return false;
+        }
+        if (Arrays.asList(artifactIds.split(",")).contains(mavenProject.getArtifactId())) {
```
martinvw:
```java
return Arrays.asList(artifactIds.split(",")).contains(mavenProject.getArtifactId())

kaikreuzer:
Closing due to inactivity.

PR #317, opened 2018-07-03
create a filter to suppress violations in git ignored files

lpapazow:
Provides a solution for #255.

With the addition of this filter, checkstyle violations for git ignored files will not be logged.

The current solution for #255 is to add supressions for each check in smarthome checkstyle/suppressions.xml file. However, the suppressions file needs to be updated every time in every repo when a new check is added or renamed. Having this filter should allow this to be avoided.

### DIFF
--- a/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/filters/GitIgnoredFilesFilter.java
+++ b/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/filters/GitIgnoredFilesFilter.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2010-2018 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.tools.analysis.checkstyle.filters;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.errors.NoWorkTreeException;
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
+
+import com.puppycrawl.tools.checkstyle.api.AuditEvent;
+import com.puppycrawl.tools.checkstyle.filters.SuppressionFilter;
+
+/**
+ * @author Lyubomir Papazov - Initial contribution
+ * 
+ *         A filter that suppresses all checks for files that are ignored by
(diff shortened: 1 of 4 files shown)


sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
```diff
@@ -12,6 +12,7 @@
  -->
 
 <module name="Checker">
+  <module name="org.openhab.tools.analysis.checkstyle.filters.GitIgnoredFilesFilter"/>
```
martinvw: Formatting looks of

custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/filters/GitIgnoredFilesFilter.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public boolean accept(AuditEvent event) {
+        String fileName = event.getFileName();
+        Path pathToFileName = Paths.get(fileName);
+
+        Path pathToFolder = pathToFileName.getParent();
+
+        Path pathToGitIgnore = pathToFolder.resolve(GITIGNORE_FILE_NAME);
+        File gitIgnoreFile = pathToGitIgnore.toFile();
+        if (gitIgnoreFile.exists()) {
```
martinvw:
```java
return !gitIgnoreFile.exists()

File: sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml (hunk 1 above)
Comment (martinvw) [reply sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml]: Nevermind

**Comment by lpapazow** [comment]
@martinvw, I changed the logic completely, can you please have another look

> I do not yet see how this would work, is the service offered so good that it can handle all the syntax of gitignore files, including question marks, *, partial and full paths?

You are right, it wouldn't have been able to handle all the syntax of .gitignore files just using the .gitignore path. Instead, I now added the `jgit` library, which can provide a list of all the files that are ignored by git.

File: custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/filters/GitIgnoredFilesFilter.java [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+import com.puppycrawl.tools.checkstyle.filters.SuppressionFilter;
+
+/**
+ * @author Lyubomir Papazov - Initial contribution
+ * 
+ *         A filter that suppresses all checks for files that are ignored by
+ *         git. In order to activate the filter, A module with the full class
+ *         name should be added within the Checker module in the rules file.
+ */
+public class GitIgnoredFilesFilter extends SuppressionFilter {
```
dstivanov: Is is possible to add some unit tests for this filter?

pom.xml
```diff
@@ -66,6 +66,7 @@
     <maven.surefire.plugin.version>2.12.4</maven.surefire.plugin.version>
     <jetty.server.version>9.3.14.v20161028</jetty.server.version>
     <saxon.version>9.1.0.8</saxon.version>
+    <jgit.version>5.1.1.201809181055-r</jgit.version>
```
dstivanov: This looks pretty much like a snapshot version. Is there an official release (which could be considered more stable)?

kaikreuzer:
Closing due to inactivity.

PR #328, opened 2018-09-25
Added a bunch of check which I now manually validate 

martinvw:
Delay on build is not significant, build times were inconsistent, but I saw a difference of around 2 minutes.

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/pmd/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/pmd/rules.xml
@@ -66,6 +66,29 @@
    <rule ref="category/java/errorprone.xml/CompareObjectsWithEquals">
     <priority>3</priority>
   </rule>
+  <rule ref="category/java/errorprone.xml/AvoidLosingExceptionInformation">
+    <priority>2</priority>
+  </rule>
+
+  <!-- https://pmd.github.io/pmd-6.2.0/pmd_rules_java_codestyle.html#avoidfinallocalvariable -->
+  <rule ref="category/java/codestyle.xml/AvoidFinalLocalVariable">
+    <priority>2</priority>
+  </rule>
+  <rule ref="category/java/codestyle.xml/ShortVariable">
+    <priority>3</priority>
+  </rule>
+  <rule ref="category/java/codestyle.xml/SuspiciousConstantFieldName">
+    <priority>2</priority>
+  </rule>
+  <rule ref="category/java/codestyle.xml/UnnecessaryConstructor">
+    <priority>2</priority>
+  </rule>
+  <rule ref="category/java/codestyle.xml/UnnecessaryModifier">
+    <priority>3</priority>
+  </rule>
+  <rule ref="category/java/codestyle.xml/UselessQualifiedThis">
+    <priority>3</priority>
+  </rule>
   
   <!-- See https://pmd.github.io/pmd-6.2.0/pmd_rules_java_multithreading.html -->
   <rule ref="category/java/multithreading.xml/DontCallThreadRun">
@@ -85,7 +108,13 @@
   <rule ref="category/java/bestpractices.xml/SystemPrintln">
     <priority>2</priority>
   </rule>
-  
+  <rule ref="category/java/bestpractices.xml/LooseCoupling">
+    <priority>2</priority>
+  </rule>
+  <rule ref="category/java/bestpractices.xml/PositionLiteralsFirstInComparisons">
+    <priority>2</priority>
+  </rule>
+
   <!-- See https://pmd.github.io/pmd-6.2.0/pmd_rules_java_design.html -->
   <rule ref="category/java/design.xml/AvoidThrowingRawExceptionTypes">
     <priority>2</priority>
@@ -102,6 +131,9 @@
   
   <!-- See https://pmd.github.io/pmd-6.2.0/pmd_rules_java_performance.html -->
   <rule ref="category/java/performance.xml/AvoidArrayLoops">
+    <priority>3</priority>
+  </rule>
+  <rule ref="category/java/performance.xml/RedundantFieldInitializer">
     <priority>3</priority>
   </rule>
 </ruleset>


martinvw:
@kaikreuzer @wborn @cweitkamp please review!

dstivanov:
Thanks for your contribution!

wborn:
I'm checking what the result is of these new checks. The `ShortVariable` check also results in warnings for short Exception variable names in catch clauses such as `e` and short lambda variable names:

selection_015

Is this an indended result @martinvw? I like these short. :-)

martinvw:
https://stackoverflow.com/questions/47135929/pmd-how-to-ignore-short-variable-names-in-lambda-expressions-only

Note that it seems like both examples of you are already supported in the newest version, i see an exclusion for labmda en for catch parameters, so maybe we should update pmd 🙂

martinvw:
Indeed see https://github.com/pmd/pmd/pull/1140

Part of pmd 6.4.0

martinvw:
@wborn I prepared a branch to update PMD and the rest, feel free to review otherwise I'll try to test it tonight

martinvw:
Okay, so do you agree with any of them? It's imho als unfair to report things while reviewing which we knew were coming but they did not know, so if you disagree with those points, I think that I can better stop reviewing ;-)

martinvw:
I filtered the unique occurrences:

```
CO*
CR
GA*
ID*
IV
L
LF
NA*
NO*
O3*
OK*
ON*
PV
S
UP*
X
Y
Z
a
ai
b
b1
b2
b3
b4
bb
bc
bd
bh
bi
br
c
c1
c2
ch
ci
cl
co
cp
cs
ct
cv
d
dc
dd
df
di
dp
dr
ds
dt
e*
e5
eb
ef
eq
ev
ex
f
fc
ft
g
ga
gs
h
hc
hd
hi
i
id*
in*
io
ip
is
it
iv
j
ja
jb
jc
je
jo
jp
jr
js
k
kc
l
lc
lm
ln
m
m2
md
ml
ms*
n
nc
ni
nl
nr
ns
nx
o
o1
o2*
o3*
ok
on
op
os
p
pf
ph
pj
pr
ps
pt
pv
pw
q
q4
qo
qt
r
rc
rd
rp
rt
s
s1
s2
sb
sd
si
sn
sp
ss
st
sw
t
th
tl
tm
to
ts
tt
tx
tz
u
uv
v
w
w2
wd
wr
ws
x
xn
xy
y
yr
z
zi
```

I marked the one which could imho be considered false positives, but the rest of this list could clearly use better names :-)

kaikreuzer:
I haven't had the time to read through all of them and I think it probably takes a while to discuss them all in the round and decide on their individual value for us. At a quick glance, I would say that the 300 warnings of RedundantFieldInitializer are really something, I do not bother about in reviews. In contrary, I think having those often makes the code easier understandable. And I very much hope that it isn't that kind of comment during reviews that make your time spend on reviews be worthwhile. The paramount goal for us should be to discover bugs (that's what we had introduced pmd for) and to have readable code. 

If all of you others agree that we should have all those checks, then be it. But always remember that we should welcome contributions and not avoid them, so the more obstacles we put in the way, the worse the situation gets for people trying to contribute.

wborn:
> And I very much hope that it isn't that kind of comment during reviews that make your time spend on reviews be worthwhile. The paramount goal for us should be to discover bugs (that's what we had introduced pmd for) and to have readable code.

With this reasoning trailing white spaces (https://github.com/openhab/static-code-analysis/pull/300) aren't bugs and also don't influence code readability. Thus also aren't worthwhile for adding checks and comments?

kaikreuzer:
I am not a strong supporter of this check and I have actually NEVER added a comment in any of my reviews about trailing white spaces.
But I think this check falls into the same category as the ones reg. tabs vs. spaces and newlines at the end of a file: They prevent diffs and merge conflicts just because of some unnoticed white spaces. Such stuff should be taken care of by the code formatter, so hopefully such checks should result in 0 warnings, whenever the code formatter has been applied.

wborn:
Although the PMD explanation of the [RedundantFieldInitializer](https://pmd.github.io/latest/pmd_rules_java_performance.html#redundantfieldinitializer) check raises a valid point:

> Java will initialize fields with known default values so any explicit initialization of those same defaults is redundant and results in a larger class file (approximately three additional bytecode instructions per field).

Nice to have... but there is more to it.


Without the check you'll end up with two initialization styles. Some people who add the redundant assignments and some who don't.

Sometimes they mix up and you end up with something like in [HeliosHandler27](https://github.com/openhab/openhab2-addons/blob/a9f97a232f7008c41747665fa0dc7d2951cc5abb/addons/binding/org.openhab.binding.helios/src/main/java/org/openhab/binding/helios/handler/HeliosHandler27.java#L108-L116). Which looks messy IMHO.

Then one day someone comes along and decides to clean up the mess. Which results in git commit review noise (similar to trailing white space).

Furthermore if you only assign non-default values they standout more:

```java
public class C {
    boolean b = false;
    int i = 12345;
    short s = 0;
    MyClass myClass = null;
}
```

```java
public class C {
    boolean b;
    int i = 12345;
    short s;
    MyClass myClass;
}
```

It's clear there is something "special" about `i`.

Also you can read/write fewer code. Which is IMHO a good thing with an already verbose language like Java.

So in the end I think it is a good example of "less is more".

kaikreuzer:
A good example of what I meant that every single check can be discussed quite a while, so we should really focus on the important ones.
In this case, I do not see the necessity to bother contributors with having to switch from one style to another - actually, I often prefer it more explicit as otherwise people might not remember whether a boolean is true or false by default... For me, it makes the code better readable and if we weren't for explicit declarations, we would anyhow do Javascript :-)

But however you put it, I do not think that the benefit we are getting from that outweighs the efforts that contributors have to do and the efforts we would have to do to change the whole code base wrt that rule initially - because that's how the SAT team so far worked: Before introducing a new rule, they made sure that the code base is in line with that, so that there are no warnings once the rule becomes active. Whatever we decide about the rules of this PR, we should do the same before switching to the new SAT version. Are you up to that?

martinvw:
> Are you up to that?

Resolved, see https://github.com/openhab/static-code-analysis/pull/331

wborn:
>  people might not remember whether a boolean is true or false by default

They have to know these defaults anyway. Other contributors will still commit code without the redundant initializations. Also their own code could one day be "cleaned up" by someone who has a preference for the other initialization style. These "cleanups" will then result in unnecessary diffs.

PR #329, opened 2018-09-26
add defualt values for forbiddenPackages and forbiddenPackageUsageCh…

lpapazow:
Add default values for `checkstyle.forbiddenPackageUsageCheck.forbiddenPackages` and
checkstyle.forbiddenPackageUsageCheck.exceptions.

This is necessary because eSH and other repos that have their own 'checkstyle.properties' currently don't have these properties defined and the build fails.

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
@@ -201,8 +201,8 @@
     </module>
     <module name="org.openhab.tools.analysis.checkstyle.ForbiddenPackageUsageCheck">
       <property name="severity" value="warning" />
-      <property name="forbiddenPackages" value="${checkstyle.forbiddenPackageUsageCheck.forbiddenPackages}"/>
-      <property name="exceptions" value="${checkstyle.forbiddenPackageUsageCheck.exceptions}"/>
+      <property name="forbiddenPackages" value="${checkstyle.forbiddenPackageUsageCheck.forbiddenPackages}" default=""/>
+      <property name="exceptions" value="${checkstyle.forbiddenPackageUsageCheck.exceptions}" default=""/>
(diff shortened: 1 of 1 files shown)


sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
```diff
@@ -201,8 +201,8 @@
     </module>
     <module name="org.openhab.tools.analysis.checkstyle.ForbiddenPackageUsageCheck">
       <property name="severity" value="warning" />
-      <property name="forbiddenPackages" value="${checkstyle.forbiddenPackageUsageCheck.forbiddenPackages}"/>
-      <property name="exceptions" value="${checkstyle.forbiddenPackageUsageCheck.exceptions}"/>
+      <property name="forbiddenPackages" value="${checkstyle.forbiddenPackageUsageCheck.forbiddenPackages}" default=""/>
```
martinvw: I don't know this syntax, but if you cannot do:

```
${checkstyle.forbiddenPackageUsageCheck.forbiddenPackages:""}
```

And this works then LGTM

lpapazow: No, that didn't work. It still results in 
>Property ${checkstyle.forbiddenPackageUsageCheck.forbiddenPackages} has not been set

exception when building eSH.

hsudbrock:
The build of the `ZigBee binding` fails due to these two properties missing (probably because the version of the SAT plugin was bumped to `0.6.1` in openHAB last week). (See this issue: https://github.com/openhab/org.openhab.binding.zigbee/issues/420).

Is it still intended that this PR is being merged, or should projects using the SAT plugin now be required to provide values for those properties?

kaikreuzer:
Closing due to inactivity.

PR #331, opened 2018-09-27
Revert new rules & bump the version of PMD / spotbugs and checkstyle

martinvw:
It does require some tests on the existing code bases

### DIFF
--- a/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/SpotBugsChecker.java
+++ b/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/SpotBugsChecker.java
@@ -27,7 +27,6 @@ import javax.xml.bind.Unmarshaller;
 
 import org.apache.maven.model.Dependency;
 import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
@@ -82,7 +81,7 @@ public class SpotBugsChecker extends AbstractChecker {
     /**
      * The version of the spotbugs-maven-plugin that will be used
      */
-    @Parameter(property = "maven.spotbugs.version", defaultValue = "3.1.7")
+    @Parameter(property = "maven.spotbugs.version", defaultValue = "3.1.6")
     private String spotbugsMavenPluginVersion;
 
     /**
@@ -94,7 +93,7 @@ public class SpotBugsChecker extends AbstractChecker {
     /**
      * The version of the spotbugs that will be used
      */
-    @Parameter(property = "spotbugs.version", defaultValue = "3.1.2")
+    @Parameter(property = "spotbugs.version", defaultValue = "3.1.7")
     private String spotBugsVersion;
 
     /**
@@ -137,7 +136,7 @@ public class SpotBugsChecker extends AbstractChecker {
     private static final String SPOTBUGS_VISITORS_PROPERTY = "spotbugs.visitors";
 
     @Override
-    public void execute() throws MojoExecutionException, MojoFailureException {
+    public void execute() throws MojoExecutionException {
         Log log = getLog();
 
         Properties userProps = loadPropertiesFromFile(SPOTBUGS_PROPERTIES_FILE);
@@ -152,7 +151,7 @@ public class SpotBugsChecker extends AbstractChecker {
         log.debug("Filter location is " + excludeLocation);
         userProps.setProperty(SPOTBUGS_EXCLUDE_FILTER_USER_PROPERTY, excludeLocation);
 
-        String visitors = getVisitorsString(spotbugsRuleset, DEFAULT_VISITORS_XML);
+        String visitors = getVisitorsString(spotbugsRuleset);
         log.debug("SpotBugs visitors " + visitors);
         userProps.setProperty(SPOTBUGS_VISITORS_PROPERTY, visitors);
         String outputDir = userProps.getProperty("spotbugs.report.dir");
@@ -192,15 +191,15 @@ public class SpotBugsChecker extends AbstractChecker {
                 pluginList.add(element);
             }
         }
-        return new Element("plugins", pluginList.toArray(new Element[pluginList.size()]));
+        return new Element("plugins", pluginList.toArray(new Element[0]));
     }
 
-    private Element createPlugin(String groudId, String artifactId, String version) {
-        return element("plugin", element("groupId", groudId), element("artifactId", artifactId),
+    private Element createPlugin(String groupId, String artifactId, String version) {
+        return element("plugin", element("groupId", groupId), element("artifactId", artifactId),
                 element("version", version));
     }
 
-    private String getVisitorsString(String externalLocation, String defaultLocaiton) throws MojoExecutionException {
+    private String getVisitorsString(String externalLocation) {
         // Get the XML file as a Stream
         InputStream stream = null;
         if (externalLocation != null) {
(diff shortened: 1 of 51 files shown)


wborn:
Thanks @martinvw! I'll give it a go and see if it fixes the `ShortVariable` warnings for variables names in and exception clauses and lambdas (https://github.com/openhab/static-code-analysis/pull/328#issuecomment-424864404).

martinvw:
Did you give it a go already?

wborn:
Not yet but I think I'll do that this evening!

martinvw:
Never mind than, I will also have some time to do it myself :-)

sat-plugin/src/main/java/org/openhab/tools/analysis/tools/SpotBugsChecker.java
```diff
@@ -82,7 +82,7 @@
     /**
      * The version of the spotbugs-maven-plugin that will be used
      */
-    @Parameter(property = "maven.spotbugs.version", defaultValue = "3.1.5")
+    @Parameter(property = "maven.spotbugs.version", defaultValue = "3.1.7")
     private String spotbugsMavenPluginVersion;
```
wborn: There's a Spotbugs Maven plugin and Spotbugs itself. 

The most recent released version of: 
* Spotbugs Maven plugin is 3.1.6
* Spotbugs is 3.1.7 (the field below, which also needs to be updated.

martinvw: This was already out-of-sync before, good that you noticed 👍 

Let me also update the documentation

wborn:
The `ShortVariable` is definitly improved with the new version. But I still see warnings for variables that I think are OK such as 'sb' for StringBuilders and `it` for iterators that are defined outside loops or `b` for byte. WDYT of the results @martinvw?

martinvw:
I'm not such a fan of the examples which seem to remain, so if we can make it the lowest priority I'm fine, can you agree with that? Both a `byte` and a `StringBuilder` also have some meaning and bigger scope. Way more than the `i` in a loop, the `e` in a catch and the `?` in lambda. 

So I personally like the result, @kaikreuzer @maggu2810 @cweitkamp what do you think?

wborn:
Here's an example report: [summary_report.html.gz](https://github.com/openhab/static-code-analysis/files/2425410/summary_report.html.gz)

> make it the lowest priority I'm fine

I always wonder what that exactly means when reading the [Guidelines](https://www.openhab.org/docs/developer/development/guidelines.html#e-logging)

> Please fix all the priority 1 issues and all issues with priority 2 and 3 that are relevant (if you have any doubt don't hesitate to ask).

Is it OK for a contributor to just fix prio 1 and skip 2, 3 all together if he/she doesn't agree with these items?

kaikreuzer:
> So I personally like the result, @kaikreuzer @maggu2810 @cweitkamp what do you think?

Is this the chance to comment on the new checks that have been introduced with #328 (as I had missed to comment there)?

I see that we are adding 1412 warnings for ShortVariable alone - and many of those I would consider false positives, or at least acceptable cases.
When we started the SAT tooling, our aim was to enforce the coding guidelines and potentially have the chance to get the SAT warnings to zero - we wanted to very conservatively add new checks that are in line with the agreed guidelines.
The issue that I see is that we are now adding thousands of warnings that we will never ever get away. This renders those warnings rather useless and they distract the focus from the real issues, so that those are not taken seriously anymore either.

> Is it OK for a contributor to just fix prio 1 and skip 2, 3 all together if he/she doesn't agree with these items?

No, this is not ok. As explained, the idea is to enforce the coding guidelines. So ALL warnings should be fixed and 2+3 priorities are not breaking the build as there can be many cases where you have false positives or at least situations that can be discussed whether the rule should apply or not.

So my clear preference is to only add new checks, if they are clearly helping to enforce the agreed guidelines (which are documented at https://www.eclipse.org/smarthome/documentation/development/guidelines.html) and we stand a chance to keep the overall number of warnings low.

martinvw:
@dstivanov I tested it and it looks all fine to me.

Note that I had to add  https://github.com/openhab/openhab-core/pull/407

wborn:
I'm happy with the other new checks @martinvw! These make the code better and so far I don't see them generating false positives. I had hoped the dependency upgrade had done the same for the `ShortVariable` check.

martinvw:
Local branch is fine, triggering rebuild

kaikreuzer:
Thanks for the ping (although my name is @kaikreuzer ;-)).

Just to recap what we discussed today: It might make sense to clearly define the meaning of the different levels of checks and I think we can find a solution that fits everyones expectation. My suggestion was:
- ERROR: Must be fixed and SAT is sure about it, so it directly fails the build.
- WARN: SAT considers it a violation of our documented guidelines. Those warning must be addressed in order for the PR to be merged. If warnings are still left, there should be a comment, which clearly explains, why this one is an exception from the rule. We must strive to have 0 warnings in our code base in the repo.
- INFO: Code style issues - nothing that really prevents a merge, but which could potentially be addressed to make the code nicer and better readable/maintainable. Nothing to be considered a violation against the documented guidelines, though. It's ok to have INFOs in the code base.

Would you all agree to those?

dstivanov:
@kaikreuzer I agree with one small remark - when a PR build is finished, the warning can be seen only in the logs, but not on the PR. That means that it will not be easily visible for the PR author to fix the issues, but instead s/he will have to wait until a reviewer comes to check the PR. Will it be a good idea to introduce new github PR check for the SAT ?

dstivanov:
@kaikreuzer I agree with one small remark - when a PR build is finished, the warning can be seen only in the logs, but not on the PR. That means that it will not be easily visible for the PR author to fix the issues, but instead s/he will have to wait until a reviewer comes to check the PR. Will it be a good idea to introduce new github PR check for the SAT ?

dstivanov:
@kaikreuzer I agree with one small remark - when a PR build is finished, the warning can be seen only in the logs, but not on the PR. That means that it will not be easily visible for the PR author to fix the issues, but instead s/he will have to wait until a reviewer comes to check the PR. Will it be a good idea to introduce new github PR check for the SAT ?

dstivanov:
@kaikreuzer I agree with one small remark - when a PR build is finished, the WARNs can be seen only in the logs, but not in the PR status. This means that it will not be easily visible for the PR author to see and fix the issues, but instead s/he might wait until a reviewer comes to check the PR. 

Will it be a good idea to introduce new github PR check for the SAT ?

kaikreuzer:
> when a PR build is finished, the WARNs can be seen only in the logs, but not in the PR status.

Our plan was to have the Jenkins PR builds to keep the SAT results as an artifact available and that this should always be linked to when reviewing a PR. If we can somehow even automatically add a comment with the link to these results, it would be imho the best.

> Will it be a good idea to introduce new github PR check for the SAT ?

What exactly do you have in mind here?

kaikreuzer:
> when a PR build is finished, the WARNs can be seen only in the logs, but not in the PR status.

Our plan was to have the Jenkins PR builds to keep the SAT results as an artifact available and that this should always be linked to when reviewing a PR. If we can somehow even automatically add a comment with the link to these results, it would be imho the best.

> Will it be a good idea to introduce new github PR check for the SAT ?

What exactly do you have in mind here?

kaikreuzer:
> when a PR build is finished, the WARNs can be seen only in the logs, but not in the PR status.

Our plan was to have the Jenkins PR builds to keep the SAT results as an artifact available and that this should always be linked to when reviewing a PR. If we can somehow even automatically add a comment with the link to these results, it would be imho the best.

> Will it be a good idea to introduce new github PR check for the SAT ?

What exactly do you have in mind here?

kaikreuzer:
> when a PR build is finished, the WARNs can be seen only in the logs, but not in the PR status.

Our plan was to have the Jenkins PR builds to keep the SAT results as an artifact available and that this should always be linked to when reviewing a PR. If we can somehow even automatically add a comment with the link to these results, it would be imho the best.

> Will it be a good idea to introduce new github PR check for the SAT ?

What exactly do you have in mind here?

kaikreuzer:
> when a PR build is finished, the WARNs can be seen only in the logs, but not in the PR status.

Our plan was to have the Jenkins PR builds to keep the SAT results as an artifact available and that this should always be linked to when reviewing a PR. If we can somehow even automatically add a comment with the link to these results, it would be imho the best.

> Will it be a good idea to introduce new github PR check for the SAT ?

What exactly do you have in mind here?

PR #335, opened 2019-01-20
Disable AboutHtmlCheck

wborn:
Removes the AboutHtmlCheck check from the rules.xml because it is no longer used now all openHAB repositories are also moving to EPL-2.0.

Fixes the following exception during builds:

```
java.io.IOException: Unable to get https://raw.githubusercontent.com/openhab/openhab2-addons/master/src/etc/about.html
```

We should create a new check for checking NOTICE files instead.


wborn:
I've removed the code and tests for this check as well now @kaikreuzer.

PR #340, opened 2019-03-02
Updated schema urls to check for openHAB schemas

Hilbrand:
With move of ESH to openhab-core schema locations of xsd files can be moved to openHAB. The schemas itself are already available on openhab website.


kaikreuzer:
Hm, sounds as if I have to do a 0.6.1 release shortly after the 0.6.0 release from last week...

PR #349, opened 2019-04-01
Fixes karafFeatureCheck

Hilbrand:
feature.xml file was moved. Updated to new location.


wborn:
The feature file moved again with https://github.com/openhab/openhab2-addons/pull/5546. So I think we should undo this PR and update the other repos accordingly? Another option could be to make the dir configurable so each repo can override the feature file location. In the future the check may also need to be updated for https://github.com/openhab/openhab2-addons/pull/5555.

PR #355, opened 2019-05-10
Fix SAT for bundles with /lib, enable spotbugs for all

### LINKED ISSUE #354: No report created when bundle has /lib
Final report is not created if  bundle has /lib and addjars-mojo is executed. The XML files are created in `/target/com.googlecode.addjars.mojo.AddJarsMojo/target/code-analysis` and the report tries to assemble it from `/target/code-analysis` (which is empty).

Can be easily reproduced by build a single bundle from openhab2-addons with/without libs (e.g. airquality and allplay). This seems to be similar to https://github.com/openhab/openhab2-addons/pull/5580

J-N-K:
Fixes #354 

- correct output dir used also in case of /lib present
- spotbugs report was not included in final report because of wrong filename

Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>


J-N-K:
The other checks suffered from the same problem, fixed that, too.

J-N-K:
@kaikreuzer This is the PR I mentioned. I think it would be great to have this in a released version.

kaikreuzer:
Ok, thanks, will put a new release on my todo list.

kaikreuzer:
Here you go: https://github.com/openhab/static-code-analysis/releases/tag/0.7.0

PR #362, opened 2019-11-02
Fix ForbiddenPackageUsageCheck false positives

wborn:
The check does not clear `importsToLineNumbers` so it would repeat any findings for subsequently analyzed files.


kaikreuzer:
@wborn Let me know when you think we should do the 0.8.0 release.

wborn:
I'd love to see a release as soon as possible so we can use it to prevent new SAT errors :bug: from entering the code base. :slightly_smiling_face:

Hilbrand:
Would be nice it include #361 in a 0.8.0 release.

kaikreuzer:
@Hilbrand Makes a lot of sense, yes 👍 .

kaikreuzer:
Done: https://github.com/openhab/static-code-analysis/releases/tag/0.8.0

wborn:
Awesome! :+1:

Issue #341, opened 2019-03-03
pom first approach


maggu2810:
On Eclipse SmartHome the tool has been enabled if there is a `META-INF/MANIFEST.MF` file present.
See: https://github.com/eclipse/smarthome/blob/f2a1f38/pom.xml#L591

As such files does not exist on a pom first approach (e.g. Maven + Bnd), I wonder if the static-code-analysis is working for such projects.
I assume your tests depends on the manifest file that is "hard coded" and present in front of the generation in front of the package phase.

Can you give me some insides and if it is not working if you plan to add support?

wborn:
Perhaps the `MANIFEST.MF` location can be made configurable?
It's still generated as `target/classes/META-INF/MANIFEST.MF`

splatch:
Its quite good question if we/you should apply any checks on generated resource. Previously it was maintained manually so at least from maintenance point of view it was important to keep proper order of imports, exports etc.
Do we care about that when we generate this metadata?

wborn:
I think you can still add the wrong content to `bnd.bnd` files which may introduce SAT issues even though the manifest is generated. @maggu2810 also [commented](https://github.com/martinvw/openhab-webui/pull/1#issuecomment-468314423) that in the future annotations may be used instead of a `bnd.bnd` file. So it might not be useful to write new SAT code for validating the `bnd.bnd` file content.

splatch:
Fair enough, then its easier to verify output than follow bnd macro language.

PR #364, opened 2019-11-05
Removed checks related to MANIFEST.MF file

Hilbrand:
With the new build system this file is generated. So all checks become obsolete.


wborn:
This seems to be related to the discussion in https://github.com/openhab/static-code-analysis/issues/341. We might still want to check for issues in manifests resulting from wrong parameters in bnd.bnd files, annotations or POM files. Because of there are so many ways that can influence the generated manifest, it may be easier to just check the generated manifest.

kaikreuzer:
@Hilbrand: @wborn seems to have a point here, so shall we simply close this PR?

Hilbrand:
In theory there might some value to it. But looking at these checks, most of these, if they would trigger, seems more of a problem in core than a user specific problem. So I'm not sure if it would be right to check in such a indirect way. I didn't see any concrete examples suggested of possible usefull checks. So I'm not sure what the actual validity is that it's useful to do the checks on the generated file.
I don't think the checks are currently run at all because the file location changed. So my guess is if we want to have useful checks we need to develop new checks or at least make sure the checks are run at all. Otherwise I think we should just remove them as it's just dead code.

kaikreuzer:
@Hilbrand & @wborn What do you suggest on how to proceed here?

Hilbrand:
This pr can be redone if needed. pr collects ~dust~ conflicts, so I'll close for now.

PR #371, opened 2020-01-14
Moved "ESH-INF/" folders to "OH-INF/" folders

cweitkamp:
- Moved `ESH-INF/` folders to `OH-INF/` folders

Related to https://github.com/openhab/openhab-core/pull/1328

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

Issue #366, opened 2019-11-19
Executing spotbugs goal results in warnings on Java 11


wborn:
Each time the SAT spotbugs goal is executed on Java 11 the following warnings are logged:

```
[INFO] --- sat-plugin:0.8.0:spotbugs (sat-all) @ org.openhab.core ---
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.sun.xml.bind.v2.runtime.reflect.opt.Injector (file:/home/wouter/.m2/repository/com/sun/xml/bind/jaxb-impl/2.2.11/jaxb-impl-2.2.11.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int)
WARNING: Please consider reporting this to the maintainers of com.sun.xml.bind.v2.runtime.reflect.opt.Injector
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[INFO] Adding dependency to findbugs:0.8.0
[INFO] Adding dependency to bug-pattern:1.2.4
[INFO] Adding dependency to spotbugs:3.1.7
[INFO] Fork Value is false
     [java] JVM args ignored when same JVM is used.
     [java] Warnings generated: 1
[INFO] Done SpotBugs Analysis....
```

It seems to be caused by JAXB being used to unmarshal the spotbugs visitors list:

https://github.com/openhab/static-code-analysis/blob/cefddfa53a0d6a28f12723a78eb3c7949eaed8a2/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/SpotBugsChecker.java#L222-L233

Upgrading JAXB/javax.activation may solve these warnings. See [Using JAXB in Java 11](https://www.dariawan.com/tutorials/java/using-jaxb-java-11/).

wborn:
After fixing the JAXB warnings in #373 the plugin will log another such warning for each execution due to the spotbugs-maven-plugin usage of the Groovy ReflectionUtils class:

```
[INFO] --- sat-plugin:0.10.0-SNAPSHOT:spotbugs (sat-all) @ org.openhab.core ---
[INFO] Adding dependency to findbugs:0.10.0-SNAPSHOT
[INFO] Adding dependency to bug-pattern:1.5.0
[INFO] Adding dependency to spotbugs:3.1.12
WARNING: An illegal reflective access operation has occurred
[...]

PR #373, opened 2020-01-18
Switch to Java 11 and upgrade JAXB

wborn:
When #371 is merged SAT won't be compatible with OH 2.5.x anymore so that would also allow us to make it Java 11 only.

Upgrading JAXB fixes the "illegal reflective access operations" warnings reported in #366.

There's still one such warning remaining due to the spotbugs-maven-plugin usage of the Groovy ReflectionUtils class.

See: https://github.com/spotbugs/spotbugs-maven-plugin/issues/132

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
 sudo: required
-dist: xenial
+dist: bionic
 
 language: java
 
-jdk:
-  - openjdk8
-  - openjdk11
+jdk: openjdk11
 
 cache:
   directories:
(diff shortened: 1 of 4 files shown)


wborn:
If we decide to keep the manifest checks (see #341, #364) we should also update the Java version in the checks/tests/configuration, e.g.:

https://github.com/openhab/static-code-analysis/blob/2e148f3e2165778e3cb45f316ecf56585cb0369b/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/ManifestJavaVersionCheck.java#L24

https://github.com/openhab/static-code-analysis/blob/2e148f3e2165778e3cb45f316ecf56585cb0369b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml#L92

PR #383, opened 2020-04-04
Don't check inner classes for missing @NonNullByDefault

### LINKED ISSUE #351: Null Annotation inner classes
I'm sure this is related to #218 but if you have an inner class - the NotNullByDefault check shouldn't be applied.

Example:
```
@NonNullByDefault
public class Stuff {
   private class HelperClass {
   }
}
```
The helper class is marked as a warning "Class should be annotated with \@NonNullByDefault" incorrectly

wborn:
As a result there will still be a warning for annotating the outer class when @NonNullByDefault is missing.
It is redundant to annotate inner classes when outer classes are annotated.
The compiler generates warnings when both outer and inner classes are annotated.

Fixes #351

---

Example:

example

### DIFF
--- a/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
+++ b/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml
@@ -230,7 +230,7 @@
     
     <module name="org.openhab.tools.analysis.checkstyle.NullAnnotationsCheck">
       <property name="severity" value="warning"/>
-      <property name="checkInnerUnits" value="true"/>
+      <property name="checkInnerUnits" value="false"/>
     </module>
 
     <!-- Rules for Java Naming Convention -->


wborn:
I will check if it still does the other checks, e.g. warning when @NonNull is used. If it no longer does that we need to modify the check to only skip the `@NonNullByDefault` check on inner classes.

wborn:
The `checkInnerUnits` seems to work as advertised in the JavaDocs and only enables/disables checking for `@NonNullByDefault` on inner classes:

https://github.com/openhab/static-code-analysis/blob/1d154f8b149eca630146502253b36e369be8837f/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/NullAnnotationsCheck.java#L54-L59

Looks like Eclipse also checks and warns for such issues this nowadays :+1: :

test

wborn:
This change will also reduce the number of SAT `NullAnnotationsCheck` findings since it will only  warn about missing `@NonNullByDefault` at most once per file.

The NullAnnotationsCheck findings are:

|Project|Before|After|Delta|
|-|-|-|-|
|openhab-core|1162|997|165|
|openhab-addons|4209|3712|497|

PR #405, opened 2021-01-14
[custom] Fix for feature check to ignore bindings with no feature file

Hilbrand:
After running this check on the current openhab-addons repo I found out the modbus has no feature file anymore. The check could not handle such cases and not be configured to ignore it. So I've changed the configuration settings, and also renamed it to better reflect what it actually does.


wborn (approved):
LGTM & thanks!

PR #427, opened 2022-01-07
Update license headers to 2022

cweitkamp:
- Update license headers to 2022

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

### DIFF
--- a/sat-plugin/src/main/java/org/openhab/tools/analysis/report/ReportMojo.java
+++ b/sat-plugin/src/main/java/org/openhab/tools/analysis/report/ReportMojo.java
@@ -1,37 +1,14 @@
 /**
- * Copyright (C) 2012-2013, Markus Sprunck
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
  *
- * All rights reserved.
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
  *
- * Redistribution and use in source and binary forms, with or
- * without modification, are permitted provided that the following
- * conditions are met:
+ * This program and the accompanying materials are made available under the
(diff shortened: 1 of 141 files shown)


sat-plugin/src/main/java/org/openhab/tools/analysis/report/ReportMojo.java
```diff
@@ -1,37 +1,14 @@
 /**
```
cweitkamp, line 1: Was this header different by intention?

wborn: Yes see https://github.com/openhab/static-code-analysis/blob/main/LICENSE

It's excluded from the check:

https://github.com/openhab/static-code-analysis/blob/7b5050dc554634e2cb0108aec041eec0c2eb80c6/tools/static-code-analysis/checkstyle/suppressions.xml#L8

kaikreuzer: I think so, yes. It seems that [this code was copied](https://github.com/openhab/static-code-analysis/pull/427/files/8bd1abd56a7558b8d0c7122ad2dee8e650ec9158#diff-5895a1cf809a55b9aeb3485f43d31d3da4e121ad3c5001ebb0898611ca3bcdefR66-R71) and is licensed under BSD, so we should keep the copyright header in place.

cweitkamp: Alright. I reverted it in 76b0970 and excluded it from further formatting.

wborn (approved):
Very good!

PR #428, opened 2022-02-11
Allow `@NonNull` on generic argument and parameter types

wborn:
When classes are annotated with `@NonNullByDefault` this annotation does not apply to generic argument and parameter types.
With these changes the `NullAnnotationsCheck` allows for `@NonNull` annotations on generic argument and parameter types.

This fixes for instance 48 `NullAnnotationsCheck` false positives in openhab-core and makes it easier to find and fix real `@NonNull` misusages.


kaikreuzer (approved):
Thanks!

PR #429, opened 2022-02-12
Fix OhInfXmlUsageCheck configurable service unused configuration false positives

wborn:
This fixes many "Unused configuration reference with uri - ..." OhInfXmlUsageCheck false positives of configurations used by configurable services.
When there are still unused configuration references, the check will now also take into account the "service.config.description.uri" properties in the generated XML of service components in the OSGI-INF dir.


kaikreuzer (approved):
Thanks!

PR #430, opened 2022-02-20
Suppress OhInfXmlUsageCheck profile unused configuration false positives

wborn:
Suppresses unused configuration reference warnings for configuration description URIs starting with "profile:".
Currently the only way to figure out what profile type UIDs are used in a bundle seems to be by instantiating each ProfileFactory.


martinvw (approved):

kaikreuzer (approved):
Thanks!

PR #433, opened 2022-02-21
Remove obsolete MANIFEST.MF and build.properties checks

wborn:
These checks have become obsolete after switching to Bnd for building bundles.
The build.properties file is no longer used and the MANIFEST.MF is generated which prevents the human errors that were found by the checks.


kaikreuzer (approved):
Makes sense, thanks.

PR #435, opened 2022-03-17
Update documentation

wborn:
* Replace ESH with openHAB
* Replace FindBugs with SpotBugs
* Fix broken links and paths


kaikreuzer (approved):
Such outdated docu - thanks for fixing it!

PR #441, opened 2022-12-20
Switch to Java 17, upgrade dependencies

wborn:
Also fixes some warnings/errors caused by the upgrades.


kaikreuzer (approved):

kaikreuzer:
@wborn Do we need a new SAT release any time soon for openHAB 4 or does it still work well with the existing one?

wborn:
The current release already works very well with Java 17.

PR #443, opened 2023-01-05
Update license headers to 2023

wborn:
I have updated the headers because it is now 2023. :slightly_smiling_face:


kaikreuzer (approved):
Thanks!

PR #453, opened 2023-08-11
Prevent warnings when running pmd goal

wborn:
Fixes the following warnings when building with Maven 3.9.x:

`[WARNING]  Parameter 'outputEncoding' (user property 'outputEncoding') is read-only, must not be used in configuration`


martinvw (approved):
LGTM

kaikreuzer (approved):

PR #458, opened 2024-01-09
Cleanup code

wborn:
This cleanup includes:

* Fix deprecations
* Fix format argument mismatches
* Fix JavaDocs
* Fix typos
* Remove redundant boxing
* Remove redundant imports
* Remove redundant modifiers
* Remove redundant toString calls
* Remove redundant variable initialization
* Use diamond operator
* Use enhanced for loops
* Use secure URLs with XML
* Use try-with-resources with Files.list
* Replace Collectors.joining with String.join
* Replace lambdas with method references


kaikreuzer (approved):
Cool, thanks.

May I ask which AI process you are using to catch and fix all those things?

wborn:
IntelliJ IDEA helps me find all these issues and it can also fix most automatically which saves a lot of time and regressions!
Usually I analyze all the code and only fix a subset of all the found issues.
It can also analyze all code using one rule which helps if you want to fix one or two problems in a lot of code like https://github.com/openhab/openhab-addons/pull/16249.
I use both Eclipse and and IntelliJ IDEA but didn't use IDEA yet with openHAB.

PR #461, opened 2024-02-29
Upgrade SpotBugs from 4.7.3 to 4.8.3 (#447)

holgerfriedrich:
Refs openhab/openhab-distro#1590

This adds support for Java file version 65 (Java 21). For processing Java 21 files, we need at least to upgrade to 4.8.0 which uses newer version of ASM dependency.

Changelog:
https://github.com/spotbugs/spotbugs/blob/master/CHANGELOG.md

In addition:
- Upgrade SpotBugs Maven plugin to 4.8.3.1.
- Allow Java11 and Java17 compilers.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -482,7 +482,7 @@
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <version>[11.0,12.0),[17.0,18.0)</version>
+                  <version>[11.0,12.0)</version>
                 </requireJavaVersion>
               </rules>
             </configuration>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -482,7 +482,7 @@
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <version>[11.0,12.0)</version>
+                  <version>[11.0,12.0),[17.0,18.0)</version>
```
wborn, line 485: It's there because we want to make sure it is released using Java 11, see https://github.com/openhab/openhab-distro/issues/1540.

```suggestion
                  <version>[11.0,12.0)</version>
```

wborn (approved):
Thanks for the upgrade!

PR #464, opened 2024-06-14
Fix dark mode scrollbars in Chrome

wborn:
This fixes the issue that when using Chrome the scrolbars are still very bright and not dark.

### Before

Screenshot from 2024-06-14 22-56-57

### After

Screenshot from 2024-06-14 22-58-01


kaikreuzer (approved):
Nice, thanks.

PR #465, opened 2024-06-14
Fix broken build

wborn:
* Use valid item type to fix the test
* Add new test for testing an invalid item type

Caused by https://github.com/openhab/website/pull/461


kaikreuzer (approved):
Thanks!

PR #478, opened 2025-01-15
Allow to configure the plugin to fail on any severity

david-m-s:
This is necessary to ease the integration to cicd with requirements to not pass if any static code violation is found, without needing to configure the checkstyle, pmd and findbugs in depth to produce only error level. It s hard to configure each individual plugin (if at all possible) and can be difficult to maintain regarding future versions.


sat-plugin/src/main/java/org/openhab/tools/analysis/report/ReportMojo.java
```diff
@@ -342,12 +364,21 @@ private int countPriority(NodeList messages, String priority) {
...
+        addLevel(failOnDebug, "@priority=3", errorPriorities);
+        String xpathExpression = "/sca/file/message[" + String.join(" or ", errorPriorities) + "]";
+        int errorCount = selectNodes(mergedReport, xpathExpression).getLength();
         if (errorCount > 0) {
-            throw new MojoFailureException(String.format(
-                    "%n" + "Code Analysis Tool has found %d error(s)! %n"
-                            + "Please fix the errors and rerun the build. %n",
-                    selectNodes(mergedReport, "/sca/file/message[@priority=1]").getLength()));
+            throw new MojoFailureException(String.format("%n" + "Code Analysis Tool has found %d error(s)! %n"
+                    + "Please fix the errors and rerun the build. %n", errorCount));
```
wborn, line 375: This message may need an update as well depending on the configuration. Users might think only fixing errors should be enough but if you configure it to fail on warnings/info this is no longer the case.

david-m-s: Updated the message to detail the levels to fix.

sat-plugin/src/main/java/org/openhab/tools/analysis/report/ReportMojo.java
```diff
@@ -115,6 +117,18 @@ public class ReportMojo extends AbstractMojo {
...
+     * Describes of the build should fail if medium priority error is found
+     */
+    @Parameter(property = "report.fail.on.warning", defaultValue = "false")
+    private boolean failOnWarning;
+
+    /**
+     * Describes of the build should fail if low priority error is found
+     */
+    @Parameter(property = "report.fail.on.debug", defaultValue = "false")
+    private boolean failOnDebug;
```
wborn, line 130: Is there a reason why you opted for calling this `debug` instead of `info`?

The output uses info for this: 

```
[INFO] --- sat:0.17.0:report (sat-all) @ org.openhab.core.thing.tests ---
[WARNING] Code Analysis Tool has found: 
 0 error(s)! 
 1 warning(s) 
 0 info(s)
```

david-m-s: No particular reason, I updated this to be "info" as it probably makes more sense.

sat-plugin/src/test/java/org/openhab/tools/analysis/report/ReportMojoTest.java
```diff
@@ -57,29 +62,32 @@ public void setUp() throws Exception {
...
+    @ParameterizedTest
+    @MethodSource("provideReportParameters")
+    public void assertReportIsCreatedAndBuildFailsAsExpected(boolean failOnError, boolean failOnWarning,
+            boolean failOnDebug, boolean buildFailExpected) {
         assertFalse(resultFile.exists());
 
-        subject.setFailOnError(true);
+        subject.setFailOnError(failOnError);
+        subject.setFailOnWarning(failOnWarning);
+        subject.setFailOnInfo(failOnDebug);
```
wborn, line 73:
```suggestion
        subject.setFailOnInfo(failOnInfo);
```

PR #483, opened 2025-04-27
Upgrade all tools

holgerfriedrich:
draft, this is for checking java11 build


holgerfriedrich:
Maybe we can do a release after this is merged?

wborn (approved):
Thanks for the upgrades!

PR #490, opened 2025-08-12
Add check for Optional used as a field type

jlaur:
This adds a check to generate a warning when `Optional` is used as a field type.

`Optional` is primarily meant as a return type, not a field type. See for example Brian Goetz, Java’s language architect, [shedding light](https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555) on Oracle’s intention with the `Optional` type.

Example code which did not generate any warnings, but hides a potential `NullPointerException` in a concurrency scenario:

```java
private Optional<String> foo = Optional.empty();

private void test()
{
    if (foo.isPresent()) {
        String bar = foo.get();
    }
}
```

Whereas this will generate a warning:

```java
private @Nullable String foo;

private void test()
{
    if (foo != null) {
        String bar = foo;
    }
}
```

and this will not:

```java
private @Nullable String foo;

private void test()
{
    String foo = this.foo;
    if (foo != null) {
        String bar = foo;
    }
}
```

This is a follow-up to the suggestion made here: https://github.com/openhab/openhab-addons/pull/19061#issuecomment-3128266716

A few @openhab/add-ons-maintainers already favors the idea, but it would be good to hear from @openhab/core-maintainers as well in case anyone thinks this approach is too drastic or are directly against it.

It should be mentioned that some add-ons already use `Optional` as a field type and thus will start having these warnings. I intend to create at least a few pull requests to address some of those.

### DIFF
--- a/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/OptionalFieldCheck.java
+++ b/custom-checks/checkstyle/src/main/java/org/openhab/tools/analysis/checkstyle/OptionalFieldCheck.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright (c) 2010-2025 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.tools.analysis.checkstyle;
+
+import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


wborn:
It's also documented in the [`Optional` JavaDoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Optional.html):

> Optional is primarily intended for use as a method return type where there is a clear need to represent "no result," and where using null is likely to cause errors. A variable whose type is Optional should never itself be null; it should always point to an Optional instance.

In IntelliJ IDEA there are also checks that show warnings when using Optional not as intended. E.g. when using it in a method parameter or in a serializable class, see:

https://www.jetbrains.com/help/inspectopedia/OptionalUsedAsFieldOrParameterType.html

PR #497, opened 2025-10-10
Introduce release process through GHA

kaikreuzer:


.github/workflows/release.yml
```diff
@@ -0,0 +1,253 @@
...
+        git add custom-checks/findbugs/pom.xml
+        git add custom-checks/pmd/pom.xml
+        git add sat-extension/pom.xml
+        git add sat-plugin/pom.xml
+        git add codestyle/pom.xml
+        git commit -m "Prepare for next development iteration: ${{ github.event.inputs.next_snapshot_version }}"
+        git push origin
+    
+    - name: Create GitHub Release
+      uses: actions/create-release@v1
```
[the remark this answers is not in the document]
kaikreuzer: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/openhab/static-code-analysis/pull/497#discussion_r2423420146)

.github/workflows/release.yml
```diff
@@ -0,0 +1,237 @@
...
+        type: string
+
+jobs:
+  release:
+    name: Release to Maven Central
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
```
holgerfriedrich:
```suggestion
      uses: actions/checkout@v6
```

holgerfriedrich: latest is [v6](https://api.github.com/repos/actions/checkout/releases/latest), should work here as well...

holgerfriedrich:
@kaikreuzer do you want to start the publishing without further measures (no dry-run or approval of the created binary?) I know this is not that easy as the version plugin changes the git.

You could also consider using `actions/upload` to upload the build artifact to GH - in case sonatype upload fails, it could be useful to have it at hand.

kaikreuzer:
> do you want to start the publishing without further measures (no dry-run or approval of the created binary?)

@holgerfriedrich What would be your suggestion? I'm open to any good approach.

holgerfriedrich:
@kaikreuzer I asked the AI to create what I described. With some more infos, the output seems reasonable:

https://github.com/holgerfriedrich/static-code-analysis/commit/43400f109452786df354501251f5b1a16a430f71

It introduces the parameter, uses it as a condition for all publish and git push sections.

Additionally, it should upload the artifacts to GH. This may need a different pattern for the respective file names.


Overall, this is the built-in and easy solution.
Benefit is that we can test the artifact.
Downside is that we need to retrigger later, setting all the parameters (version, next version) again and correctly.
I order to get around this, we would need to use those additional actions built specifically to allow a user interaction during between steps of a build.

holgerfriedrich:
@kaikreuzer almost got the first part running, see https://github.com/holgerfriedrich/static-code-analysis/commits/pr-dryrun
Upload still does not pick artefacts.

.github/workflows/release.yml
```diff
@@ -0,0 +1,259 @@
...
+    
+    - name: Update version to release version
+      run: |
+        mvn versions:set -DnewVersion=${{ github.event.inputs.release_version }}
+        mvn versions:commit
+    
+    - name: Update parent POM versions in modules
+      run: |
+        # Update parent version in all child POMs
+        for pom in $(find . -name pom.xml -not -path "*/target/*"); do
```
holgerfriedrich:
```suggestion
        for pom in $(find . -name pom.xml -not -path "*/target/*" -not -path "*/src/test/resources/*"); do
```

holgerfriedrich:
Did you check the GPG key stored in secrets?
It should be armored, i.e. start with `-----BEGIN PGP PRIVATE KEY BLOCK-----`

kaikreuzer:
Thanks @holgerfriedrich, it seems I indeed had a wrong format of the key.
Now publishing works, the pipeline now only fails because it tries to do a commit to `main`: https://github.com/openhab/static-code-analysis/actions/runs/20249411534/job/58137742534

I'd like to grant the github action an exception on the branch protection rule, but I couldn't yet figure out, how to do that...

kaikreuzer:
@holgerfriedrich It finally worked: https://github.com/openhab/static-code-analysis/releases/tag/v0.18.0

PR #562, opened 2026-05-31
Fix item type validation test and Java 21 deprecations

wborn:
Generate the expected invalid item type message from the current thing description schema instead of hard-coding the full item type enumeration. This keeps the test stable when new item types are added to the XSD.

Also replace deprecated Java 21 API usage for Locale and URL creation.


wborn:
This PR will fix the build @openhab/sat-maintainers ✅

kaikreuzer (approved):
Thanks!

